PR #27, opened 2026-07-20
fix(k3s): back the guest data dir with a volume even without persistence

jleni:
Second half of the 2026-07-20 `ci-k3s-kunobi` incident (#25 was blast-radius + forensics; this is the crashloop itself).

**Root cause (cross-verified Claude + codex, adjudicated against upstream k8s source):** without persistence the whole k3s data dir — including the nested containerd — lives in the guest container's overlayfs rw layer. containerd can't run its overlayfs snapshotter on top of overlayfs, so CRI runtime-ready takes ~25–35s; node registration is gated on it and races the kubelet's **fatal** CSINode-init window, which on k8s ≤1.32 is only **~23s** of backoff (`csi_plugin.go`: Steps 6 / 15ms / ×6 — matches our observed ~26s panics exactly). And because each container restart cold-starts from a fresh rw layer, one lost race replays identically forever — terminal CrashLoopBackOff.

**Change:** mount a pod-level `emptyDir` at `/var/lib/rancher/k3s`:
- server: only when no persistence is configured (with persistence the `volumeClaimTemplates` PVC keeps owning the `data` name — pinned by test);
- agent: always (agents have no PVC variant).

Effects: containerd gets a real node filesystem (overlayfs snapshotter works → runtime-ready in seconds), and restarts resume with certs/DB/images intact — a lost first boot self-heals on restart #2 instead of crashlooping.

**Companion (margin half):** bump the pool's guest k3s to ≥ v1.33 — upstream widened the CSINode-init backoff ~23s → ~140s in release-1.33. That's a pool-config change in tenant-int-pro.

New test: `test_no_persistence_data_dir_backed_by_emptydir` (server emptyDir + mount, no shadowing under persistence, agent emptyDir + mount). 642/642 tests pass, clippy clean.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2656,7 +2656,7 @@ dependencies = [
 
 [[package]]
 name = "kobe-operator"
-version = "0.33.0"
+version = "0.34.0"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -2716,7 +2716,7 @@ dependencies = [
 
 [[package]]
 name = "kobectl"
-version = "0.33.0"
+version = "0.34.0"
 dependencies = [
  "anyhow",
  "chrono",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,7 +9,7 @@ resolver = "3"
 # CLI so they always publish in lockstep. `just bump X.Y.Z` rewrites this; the
 # version-consistency gate enforces tag == this == Chart.yaml at release time.
 [workspace.package]
-version = "0.33.0"
+version = "0.34.0"
 edition = "2024"
 rust-version = "1.94.1"
 license = "Apache-2.0"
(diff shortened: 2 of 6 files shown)


jleni:
Added a second commit per review discussion: kobe now *knows* about the version hazard instead of leaving it tribal knowledge — a per-reconcile WARN on pools whose guest k8s is ≤1.32 (naming the ~23s vs ~140s CSINode window and the ≥v1.33 remediation) + a `kobe_pool_guest_version_advisory{advisory="narrow_csinode_window"}` 0/1 gauge for dashboards/alerts, in the style of the #190 guardrail. The emptyDir data dir from the first commit is the workaround that keeps ≤1.32 pools functional (self-healing restarts). 643/643 tests, clippy clean.

PR #45, opened 2026-07-24
refactor(demo): install kobe from the official chart, drop the umbrella

0xPxt:
## What's New

The demo now installs kobe the way you would in production — the official chart plus a couple of custom resources — instead of a bespoke demo chart. What `./demo up` does now:

```bash
helm install kobe ./charts/kobe -f values.yaml   # operator, clean
kubectl apply -f manifests/kobe/                  # ClusterPool + AccessPolicy
```

Previously an umbrella chart wrapped the official `charts/kobe` chart just to add a `ClusterPool` and `AccessPolicy`, and `./demo up` required a Docker Hub pull secret.

## Technical Details

- Deleted the `demo/_shared/chart` umbrella (chart-of-charts).
- `values.yaml` is minimal — `replicas: 1` + `operatorNamespace: kobe-system`; the chart already defaults `imagePullSecrets: []` and etcd/kine/serviceMonitor/ingress off.
- `ClusterPool` and `AccessPolicy` are standalone manifests applied with `kubectl`; the SSH key is injected into the AccessPolicy via `yq`.
- Removed the pull-secret command/env entirely and the vendored-tgz `refresh` machinery; `./demo down` now deletes the CRs too.
- Both the Hetzner and Exoscale demos move to this flow; exoscale keeps a standalone local-path manifest for its missing default StorageClass.
- Both READMEs synced.

## Testing

Ran the new `./demo up` end-to-end on a fresh kind cluster: the operator installs from `charts/kobe` (public image, no pull secret), the CRs apply, the pool warms to 2/2 Ready, exit 0. Also verified `down`, `lint`, and `template`.


demo/_shared/lib.sh
```diff
@@ -443,7 +430,9 @@ Shared subcommands:
   patch-lease    Rewrite a leased kubeconfig to use https://localhost:$TLS_PORT
                    Usage: ./demo patch-lease [LEASED_KUBECONFIG_PATH]
   status         kubectl get clusterpool,clusterinstance,clusterlease
-  pull-secret    Create regcred docker-registry secret in $NAMESPACE
+  pull-secret    (optional) Create regcred docker-registry secret in $NAMESPACE
```
jleni: lets just remove this to make things much simpler.. it will never be necessary

jleni (changes requested):

PR #46, opened 2026-07-24
ci: publish the kobe Helm chart as an OCI artifact on a helm_v* tag

0xPxt:
## What's New

Publishes the `kobe` Helm chart as an OCI artifact, so it can be installed directly without cloning the repo:

```bash
helm install kobe oci://registry-1.docker.io/zondax/kobe --version 0.0.1 -f values.yaml
```

## Technical Details

- New standalone workflow `.github/workflows/publish-chart.yml` triggered **only** on `helm_v*` tags — decoupled from the operator's `v*` release tags, so chart releases are an explicit, separate action and never rebuild/re-push the image. (`v*` globs don't match `helm_v*`, so `ci.yml` is untouched at runtime.)
- **Version match:** asserts the tag version (`helm_v0.0.1` → `0.0.1`) equals `version:` in `charts/kobe/Chart.yaml` before publishing.
- **Operator image check:** asserts `zondax/kobe-operator:<appVersion>` exists in the registry before publishing, so a chart release can't point at an operator version that was never built.
- Reuses the existing `DOCKERHUB_USER`/`DOCKERHUB_TOKEN`: `helm registry login` → `dependency build` → `package` → `helm push … oci://registry-1.docker.io/zondax`.
- Credentials and the tag name are passed via `env:` and referenced quoted — no untrusted input reaches the `run:` block.

**To cut a chart release:** set `charts/kobe/Chart.yaml` `version:` to `X.Y.Z`, then push tag `helm_vX.Y.Z`.

## Still to confirm

- **Docker Hub repo.** `zondax/kobe` doesn't exist yet — the first push creates it **private**, so it must be flipped to **public** once, and the CI token needs rights to push a new repo under `zondax`.

### DIFF
--- a/.github/workflows/publish-chart.yml
+++ b/.github/workflows/publish-chart.yml
@@ -1,0 +1,62 @@
+name: Publish Helm Chart
+
+# Chart releases ride their own tag, decoupled from the operator's v* release
+# tags: push a `helm_v<chart-version>` tag (e.g. helm_v0.0.1) to publish the
+# chart. The tag version must match `version:` in charts/kobe/Chart.yaml, so the
+# published OCI version and the chart's declared version can never drift.
+on:
+  push:
+    tags: ["helm_v*"]
+  workflow_dispatch:
+
+concurrency:
+  group: publish-chart-${{ github.ref }}
+  cancel-in-progress: false
+
+jobs:
+  publish-chart:
+    runs-on: kunobi-runners
+    timeout-minutes: 15
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: azure/setup-helm@v4
+
+      - name: Assert tag version matches Chart.yaml
+        if: startsWith(github.ref, 'refs/tags/helm_v')
+        env:
+          REF_NAME: ${{ github.ref_name }}
+        run: |
+          tag_version="${REF_NAME#helm_v}"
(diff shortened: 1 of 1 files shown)


jleni:
Merging this as-is to start publishing charts, with a refactor to follow. Context from the 0.36.0 release: the chart now tracks the operator version (#50) — `charts/kobe/Chart.yaml` `version:` == `appVersion:` == the workspace version, gated by `scripts/check-version-consistency.sh`, and `just bump` moves all three. So the first chart release off this workflow is `helm_v0.36.0`, not `helm_v0.0.1` as in the description's example.

That makes the version-match assertion here trivially satisfied, and the "operator image must exist" check a belt-and-braces version of what will become a real dependency edge.

**Planned refactor:** fold this into `ci.yml`'s `v*` release flow as a job with `needs: [version-consistency, publish]` and retire the `helm_v*` trigger. The chart and the image then always ship together from one tag — no second tag to remember, no drift between an appVersion and an image that was never built, and the registry probe becomes unnecessary because `publish` produced the image in the same run. Keeping this standalone workflow in the meantime means chart releases work today without touching the release path mid-release.

**Still blocking a first chart push** (unchanged from the description): `zondax/kobe` does not exist on Docker Hub, the first push creates it **private**, and `DOCKERHUB_TOKEN` needs rights to create a repo under `zondax`. Worth creating it public up front rather than pushing and flipping visibility after — a private chart repo fails closed for every consumer, and the failure looks like an auth error rather than a visibility one.

PR #50, opened 2026-07-26
chore(release): cut 0.37.0 and version the chart with the operator

jleni:
## Summary

- cut **0.37.0**: workspace version, `Chart.yaml` `appVersion` **and** `Chart.yaml` `version` (via `just bump 0.37.0`)
- `just bump` now rewrites the chart's `version:` too; the version gate asserts it

## Why 0.37.0 and not 0.36.0

The `v0.36.0` tag was pushed at f56f7ed (the #49 merge) without a preceding version bump, so the release gate failed:

```
version consistency FAILED for tag 0.36.0:
  - workspace version '0.35.0' != tag version '0.36.0'
```

Nothing was published for 0.36.0 — the crates.io job failed on the gate, CI's `version-consistency` failed so `release-cli` was skipped and the image push never ran, and every package publisher (apt/aur/choco/winget/brew/scoop) was cancelled. The GitHub release for it has zero assets.

The number is skipped rather than re-cut because that release was also created **by hand**, and per the trigger comment in `ci.yml` this repo has immutable releases enabled — a hand-published release can lock both its assets and its tag, which is exactly the state `release-cli` cannot upload into. Skipping to 0.37.0 avoids tag surgery entirely.

## Chart versioning

The chart now ships from the same `v*` tag as the operator (see #51), so `version:` == `appVersion:` == the workspace version. One-time jump for the chart: 0.21.19 → 0.37.0. Monotonic, so helm accepts it, and nothing pins the chart by version today — int-pro installs it by path from its GitRepository.

A chart-only fix is therefore a patch release of the whole thing, **not** an `X.Y.Z-1` suffix: semver reads `-1` as a *prerelease* of X.Y.Z, so it would sort below the release it means to fix, and helm hides prereleases from search and dependency resolution without `--devel`. Build metadata (`X.Y.Z+1`) is no better — ignored for precedence, so `helm upgrade` would not treat it as newer, and OCI tags cannot carry `+`.

## Validation

- gate passes in internal and tag mode (`tag 0.37.0 == workspace == Chart.yaml version + appVersion`)
- gate fails as intended on a chart-version mismatch and on a wrong tag
- `just bump 0.37.0` is idempotent and sets all three fields

## Release procedure after merge

Cut from `main` with `just release` — **tag push only, no hand-created GitHub release**. `_release-rust.yml` creates the release as a draft, attaches the signed assets, and publishes it, and that publish event is what fires the crates.io and package-publish cascade.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2656,7 +2656,7 @@ dependencies = [
 
 [[package]]
 name = "kobe-operator"
-version = "0.35.0"
+version = "0.37.0"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -2717,7 +2717,7 @@ dependencies = [
 
 [[package]]
 name = "kobectl"
-version = "0.35.0"
+version = "0.37.0"
 dependencies = [
  "anyhow",
  "chrono",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,7 +9,7 @@ resolver = "3"
 # CLI so they always publish in lockstep. `just bump X.Y.Z` rewrites this; the
 # version-consistency gate enforces tag == this == Chart.yaml at release time.
 [workspace.package]
-version = "0.35.0"
+version = "0.37.0"
 edition = "2024"
 rust-version = "1.94.1"
 license = "Apache-2.0"
--- a/charts/kobe/Chart.yaml
+++ b/charts/kobe/Chart.yaml
@@ -4,8 +4,8 @@ description: >-
   Kubernetes operator for instant cluster provisioning via warm pools.
   Supports multiple backends: k3s, k0s, CAPI, and kobe-sync.
 type: application
-version: 0.21.19
-appVersion: "0.35.0"
+version: 0.37.0
+appVersion: "0.37.0"
 keywords:
   - kubernetes
   - operator
--- a/justfile
+++ b/justfile
@@ -146,8 +146,12 @@ coverage-open:
 clean:
     cargo clean
 
-# Bump the release version across the workspace + Chart.yaml appVersion in
-# lockstep, then verify consistency. Usage: just bump 0.32.0  (or 0.32.0-rc.1)
+# Bump the release version across the workspace + Chart.yaml version and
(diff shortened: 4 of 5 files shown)


jleni:
Scope grew slightly after discussion — the chart now ships on the operator's version line, published from the same `v*` tag:

- `charts/kobe/Chart.yaml` `version:` 0.21.19 → **0.36.0** (one-time jump onto the operator line; monotonic, and nothing pins the chart by version today — int-pro installs it by path from its GitRepository)
- `just bump X.Y.Z` rewrites the chart's `version:` as well as `appVersion:`
- `scripts/check-version-consistency.sh` gates the chart `version:` too, so `tag == workspace == Chart.yaml version + appVersion`

A chart-only fix is now a patch release of the whole thing, **not** an `X.Y.Z-1` suffix: semver reads `-1` as a *prerelease* of X.Y.Z, so it would sort below the release it fixes and helm hides prereleases from search/dependency resolution without `--devel`. Build metadata (`X.Y.Z+1`) is no better — ignored for precedence, so `helm upgrade` wouldn't treat it as newer, and OCI tags can't hold `+`.

Verified locally: gate passes in both internal and tag mode, fails as expected on a chart-version mismatch and on a wrong tag; `just bump 0.36.0` is idempotent and sets both fields.

The demo keeps its vendored `kobe-0.21.19.tgz` snapshot — internally consistent (tarball, `KOBE_VERSION` default and the umbrella's dependency pin all agree), and #45 deletes that machinery. Until then, `./demo refresh` repackages as `kobe-0.36.0.tgz` and needs `KOBE_VERSION` set to match.

Follow-up, per the discussion on #46: merge that workflow as-is (it only fires on `helm_v*`), then refactor chart publishing into the `v*` release flow as a job with `needs: [version-consistency, publish]`, which turns its "operator image exists in the registry" probe into a real dependency edge and retires the separate tag.

jleni:
Superseded by #52 — renaming the branch to `chore/release-0.37.0` orphaned this PR. Same commit, retargeted.

Issue #36, opened 2026-07-20
conformance: vkobe legs — pool creates no members in the kind harness (ready=0 creating=0)


jleni:
Split from #34. After the five environment/harness layers were fixed in #35 (kind multi-arch import, missing kubectl, recycle-assertion name race, kobectl crate move, missing C toolchain), the vkobe smokes finally run — and expose a real behavior: the `e2e-vkobe-etcd` / `e2e-vkobe-kine-sqlite` pools sit at `ready=0 leased=0 creating=0 queue=0` for 8+ minutes after operator install. The operator reconciles the pools (the #190 limits warning fires for them) but never creates a member; no 'creates paused: datastore degraded' lines are visible in the failure tail.

Evidence: run 29764413062 (vkobe-etcd, vkobe-kine), kind-logs artifacts `kind-logs-29764413062-vkobe-*` (operator logs + cluster dump).

Since vkobe is **on hold** (see how-it-works/runtime-strategy), #35 marks the vkobe legs `continue-on-error` — still executed and visible per-leg, but not failing the nightly/tag run — until this is picked up. The k3s (production) leg is green and remains run-failing.

emmanuelm41:
Picked this up. The behaviour is real and reproducible, and the root cause turned out to be a product bug rather than anything vkobe-specific.

**Why the pools never create a member.** All four vkobe pools declare `scaling.minReady: 0`. The scale-up gate in `compute_pool_actions` is:

```rust
if counts.ready + counts.creating < warm_target && total < max_clusters {
```

`warm_target` comes from `min_ready`, so with `minReady: 0` that reads `0 < 0` — permanently false. No `PoolAction::Create` is ever emitted. Nothing in the lease path compensates either: `reserve_ready_instance` only ever hands out an already-`Ready` instance, it never creates one. So the pool sits at `ready=0 leased=0 creating=0` indefinitely, exactly as reported.

Queue depth *was* being counted, but only *after* `compute_pool_actions` had already run, and only to fill the `queueDepth` status field and metric. It never reached the scale-up decision. `scaleUpThreshold` is destructured into `_scale_up_threshold` and never read at all, while still being advertised in the CRD and reported back through the API — so a scale-to-zero pool looked configurable but could not work.

Your two observations both hold up, incidentally: the #189/#190 limits warning is a bare `warn!` with no gating, and the absence of any "creates paused: datastore degraded" line is consistent with that block never engaging. Neither was the blocker.

**Fixes, split in two:**

[...]

PR #57, opened 2026-08-07
fix(pool): let queued claims scale up a scale-to-zero pool

emmanuelm41:
## Problem

A pool with `scaling.minReady: 0` can never provision anything.

The scale-up gate in `compute_pool_actions` is:

```rust
if counts.ready + counts.creating < warm_target && total < max_clusters {
```

`warm_target` comes from `min_ready`, so with `minReady: 0` the comparison is `0 < 0` — permanently false. No `PoolAction::Create` is ever emitted. The lease path doesn't compensate: `reserve_ready_instance` only hands out an already-Ready instance, it never creates one. So claims queue, sit there, and expire on `queueTimeout`.

Queue depth *was* being counted — but at `controllers/profile.rs:1071`, **after** `compute_pool_actions` had already run at `:847`, and only to populate the `queueDepth` status field and metric. It never reached the scale-up decision.

## Fix

Carry queue depth on `PoolState` and add it to the warm target. Queued claims are demand the standing warm target doesn't cover, so they raise it:

```rust
let warm_target = warm_target + state.queue_depth;
```

Bounds are unchanged — `max_clusters` and `MAX_BURST` still cap the burst, so a queue spike can't become a create storm, and a claim an existing Ready cluster will serve doesn't provision a redundant one.

The count moves into `build_pool_state`, and the later use site reads it back off `PoolState`. The status/metric and the scale-up decision are now the same number instead of two independent LISTs, and the net LIST count per reconcile is unchanged (no early returns sit between the two points).

## Docs corrected

Two field descriptions documented behaviour the code didn't have:

- **`scaleUpThreshold`** — never read. It's destructured into `_scale_up_threshold` and discarded, while being advertised in the CRD and reported back through the API (`api/routes.rs:513`). Now documented as reserved/no-effect rather than removed, so existing manifests stay valid.
- **`minReady`** — silently replaces `size`. This is what makes a `size: 3` + `minReady: 0` manifest look like it keeps three clusters warm when it keeps none. (Same confusion #56 fixes in the README, from the docs side.)

## Tests

TDD — the two scale-to-zero tests were confirmed red before the one-line fix and green after:

- `queued_claims_scale_up_a_scale_to_zero_pool` — two queued claims provision two clusters
- `queued_claims_respect_max_clusters` — `maxClusters: 1` caps a 10-deep queue at one
- `queued_claim_does_not_double_provision_against_ready_capacity` — a Ready cluster already covers the claim
- `scaling_min_ready_overrides_spec_size_for_warm_target` — pins the `size`-is-ignored invariant
- `fixed_pool_without_scaling_fills_toward_spec_size` — the converse, for pools with no `scaling` block

Full workspace suite green (663 operator tests, up from 658), `cargo fmt --check` and `cargo clippy -- -D warnings` clean.

## Relation to #36

This is the root cause behind the vkobe conformance legs sitting at `ready=0 leased=0 creating=0` — those pools declare `scaling.minReady: 0`, so they were never going to create a member.

It isn't the whole of #36: `hack/test-smoke.ts` waits for `ready > 0` *before* leasing, so it never generates the queue demand that now drives scale-up. The harness config and CI-matrix half is left for a follow-up PR.

Refs #36

### DIFF
--- a/charts/kobe/crds/clusterpools.yaml
+++ b/charts/kobe/crds/clusterpools.yaml
@@ -707,7 +707,15 @@ spec:
                     type: integer
                   minReady:
                     default: 1
-                    description: Minimum number of warm (ready) clusters to maintain.
+                    description: |-
+                      Minimum number of warm (ready) clusters to maintain.
+
+                      Replaces [`ClusterPoolSpec::size`] entirely — once a `scaling`
+                      block is present, `size` is never read. Set to 0 for a
+                      scale-to-zero pool: nothing is kept warm, and clusters are
+                      provisioned on demand as claims queue (bounded by
+                      `max_clusters`), at the cost of paying full provisioning
+                      latency on every claim.
                     format: uint32
                     minimum: 0.0
                     type: integer
@@ -725,7 +733,13 @@ spec:
                     type: string
                   scaleUpThreshold:
                     default: 0
-                    description: Scale up when ready clusters fall to this threshold.
+                    description: |-
+                      Reserved; currently has no effect.
+
+                      Scale-up is driven by `min_ready` plus the pool's queue depth,
+                      not by a separate threshold — the pool creates whenever
+                      `ready + creating` falls below that target. The field is kept
+                      so existing manifests that set it stay valid.
                     format: uint32
                     minimum: 0.0
                     type: integer
--- a/src/controllers/profile.rs
+++ b/src/controllers/profile.rs
@@ -577,6 +577,68 @@ mod cluster_instance_tests {
         );
     }
 
+    /// Queue depth must count *unmet* demand only.
+    ///
+    /// A `Pending` lease that already carries a `clusterName` has had an
+    /// instance reserved for it — the lease controller recognises that
+    /// state and repairs it to `Bound`. Counting it as queued demand
+    /// during the reserve → patch-status window makes the pool
+    /// provision a replacement for a claim that is already served.
+    #[tokio::test]
+    async fn queue_depth_excludes_pending_leases_that_already_hold_a_cluster() {
+        let (ctx, server) = test_profile_context().await;
+
+        Mock::given(method("GET"))
+            .and(path(
+                "/apis/kobe.kunobi.ninja/v1alpha1/namespaces/test-ns/clusterinstances",
+            ))
+            .respond_with(ResponseTemplate::new(200).set_body_json(
+                crate::testutil::k8s_list_response(Vec::<serde_json::Value>::new()),
+            ))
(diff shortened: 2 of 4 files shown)


emmanuelm41:
Ran this through an independent cross-model review (a Codex leg and a separate Claude leg, generated blind from the same frozen diff, neither seeing the other). **Both independently found the same two blocking defects**, which I then reproduced with tests before fixing.

### 1. Fixed pools were silently turned into demand-scaled pools

`+ queue_depth` was applied *after* the scaling/fixed branch, so a pool with no `scaling` block got a warm target of `size + queue_depth` and provisioned toward its internal `size + 10` ceiling. Reproduced — a `size: 1` fixed pool with a queue emitted:

```
got [Create("pool-test-profile-2"), Create("pool-test-profile-3")]
```

The bad part is the asymmetry: fixed pools have **no scale-down path** (`scale_down_after` is `None`), so anything created above `size` on a queue spike is held forever with nothing to trim it. That was an undisclosed and untested behavior change for every existing non-scaling pool — and my "Bounds are unchanged" claim glossed over it.

Demand now applies only when the user opted into `scaling`.

### 2. Scale-down could delete the cluster a queued claim was counted against

Scale-up counts a queued claim as covered by an existing Ready cluster and emits no Create for it — but scale-down still measured excess against `min_ready` alone. So a scale-to-zero pool with one idle-past-`scaleDownAfter` Ready cluster and one queued claim produced, in a single reconcile, no Create **and**:

```
got [Delete("pool-test-profile-1")]
```

The claim loses the instance about to serve it, pays a full cold start, and the Delete races the lease controller reserving it. Scale-down now protects `min_ready + queue_demand`.

My `queued_claim_does_not_double_provision_against_ready_capacity` test only asserted "no Create" against a *fresh* Ready cluster, so it could never have caught this.

### 3. A claim of mine that was wrong

"No early returns sit between the two points" is false — `build_pool_state` bails early when the ClusterInstance LIST fails, so `queueDepth` reports 0 for that reconcile. It's a degraded pass either way and self-corrects, but rather than assert it away I've documented it where the value is consumed.

### Also noted, not changed here

The reviews flagged that `.owns(leases)` requires an OwnerReference, and API-created leases carry labels but no owner ref — so a new lease may not promptly wake the pool controller, leaving demand to be noticed on the periodic requeue. That's a pre-existing wiring question rather than something this PR introduces, and it deserves its own change (a `watches` mapping from `ClusterLease.spec.poolRef`) rather than being bolted on here. Worth tracking separately.

Both additions now use `saturating_add`. 665 tests green, fmt and clippy clean.

Issue #31, opened 2026-07-20
conformance: add vcluster leg to the CI matrix


jleni:
Split from #10 (deliberately deferred — hardening the working k3s path first).

The vcluster backend has no e2e harness pool yet, so it is a documented exemption in `backend_conformance_coverage_is_classified` (crd/profile.rs). Before vcluster carries production load it must join the conformance matrix:

- add a vcluster pool to `hack/e2e.ts`
- add a `just test-smoke-vcluster` recipe
- add the matrix leg in `.github/workflows/ci.yml` and flip its classification in the enforcement test

Same story applies to k0s and capi (capi additionally needs a CI infrastructure provider).

emmanuelm41:
Status check against current `main`: one of the three checklist items has already shipped, so the body's premise ("no e2e harness pool yet") is now out of date.

| Item | Status |
|---|---|
| vcluster pool in `hack/e2e.ts` | **done** — `e2e-vcluster` + `e2e-vcluster-bootstrap` landed in #81 |
| `just test-smoke-vcluster` | not done |
| matrix leg in `ci.yml` | not done |
| flip the classification | not done |

The stale part is also load-bearing: the exemption comment in `backend_conformance_coverage_is_classified` still says "vcluster: no e2e harness pool yet", which hasn't been true since #81.

**A dependency that wasn't noted on either issue:** the vcluster pools carry the same `scaling.minReady: 0` as the vkobe ones, so a `test-smoke-vcluster` built on `hack/test-smoke.ts` would have hit #36's wedge on day one — the smoke waits for `ready > 0` before leasing, and a scale-to-zero pool never warms. So #36 was effectively a prerequisite for this.

#59 does both together: the on-demand smoke path, plus the recipe, the matrix leg, and the classification flip.

While in there I also tightened the classification test itself. It asserted `ci.contains(leg)` against the **whole file** — but every backend name appears somewhere in ci.yml's prose, so vcluster was passing that check the entire time it had no leg at all. It now reads the matrix `backend:` declaration specifically, and additionally requires a `case` arm that runs a smoke recipe (otherwise a leg would reach the `unknown backend` guard at runtime). Both failure modes are mutation-checked.

**k0s is in the same half-state** and stays exempt for now: the harness has an `e2e-k0s` pool, but nothing drives it. Worth its own follow-up — the remaining work there is just a recipe and a leg, same shape as this.

emmanuelm41:
All four checklist items are on `main` as of #59:

[...]

PR #59, opened 2026-08-07
ci(conformance): add the vcluster leg and unwedge the on-demand smokes

emmanuelm41:
> **Stacked on #57** — base is `fix/scale-to-zero-pools`, not `main`. The on-demand smoke path only works once queued claims drive scale-up. Merge #57 first; this will retarget to `main` automatically.

Closes the CI half of #36 and the remaining checklist of #31.

## Why the vkobe legs were wedged (#36)

The pools are scale-to-zero (`scaling.minReady: 0`), but `hack/test-smoke.ts` waits for `ready > 0` **before** requesting a lease. Nothing was ever going to make those pools warm, so that wait could only time out — the reported `ready=0 leased=0 creating=0 queue=0` was the pool behaving exactly as configured, and the smoke driver asking it for something it was never going to give.

With #57, the lease request itself is what provisions a scale-to-zero pool. So the driver gains `POOL_ON_DEMAND`, which skips the warm-capacity wait and goes straight to leasing; the vkobe recipes set it, with a lease wait long enough to cover a cold provision.

**Why not just set `minReady: 1`?** `e2e.ts up` applies *every* pool in *every* leg. Warming them all would make each leg pay to provision the backends it isn't testing — on a kind node that's a lot of CPU for no signal. Keeping them scale-to-zero means only the pool under test provisions, and it exercises the on-demand path end to end in real CI as a side effect.

## vcluster leg (#31)

The harness has had `e2e-vcluster` / `e2e-vcluster-bootstrap` pools since #81, but nothing drove them: no recipe, no matrix leg. They're scale-to-zero too, so they'd have hit the identical wedge on day one — that dependency wasn't noted on either issue.

- `just test-smoke-vcluster` (on-demand, generous lease wait — the Helm install makes a cold start slower)
- `vcluster` added to the nightly matrix + its `case` arm
- vcluster flipped from documented-exemption to conformance-covered

## Drift guard tightened while I was in it

`backend_conformance_coverage_is_classified` asserted `ci.contains(leg)` against the **whole file**. Every backend name appears somewhere in ci.yml's prose, so a backend could be classified as conformance-covered while having no leg at all — which is precisely the state vcluster was in. The check was passing for a backend that wasn't running.

It now:
1. reads the matrix `backend:` declaration specifically, and
2. additionally requires a `case` arm that runs a smoke recipe, so a leg can't be added to the matrix without a runner (it would otherwise fall through to `unknown backend` at runtime).

Both failure modes mutation-checked: removing `"vcluster"` from the matrix, and removing its `case` arm, each fail the test. Under the old assertion, neither did.

## Deliberately not included

The vkobe legs stay `continue-on-error`. This PR makes them able to *provision*; whether the vkobe guests then reach Ready in kind is untested, and vcluster has never run in CI at all. Flipping that gate belongs in a follow-up once a nightly proves them green — turning it off now would just trade a known-red signal for an unknown-red one.

`k0s` also stays exempt: the harness has an `e2e-k0s` pool, but no smoke recipe drives it. The stale comment claiming it has no pool is corrected.

## Verification

`cargo test --workspace` green (663), `cargo fmt -- --check` and `cargo clippy -- -D warnings` clean, and `hack/test-smoke.ts` type-checks under bun.

The kind harness itself can only be validated by a nightly/dispatch run — worth dispatching one against this branch before merge.

Refs #36, #31

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -171,9 +171,11 @@ jobs:
   # push to main and on release tags, so a conformance regression is caught
   # before (main) or at (tag) the release boundary instead of a day later.
   # The full matrix (k3s + both vkobe datastores) stays nightly + manual; the
-  # vkobe legs run the same gate across etcd + kine/sqlite. vcluster and capi
-  # have no e2e harness/pool yet — tracked as follow-up before they join the
-  # matrix; `backend_conformance_coverage_is_classified` (crd/profile.rs)
+  # vkobe legs run the same gate across etcd + kine/sqlite, and vcluster joins
+  # nightly against the upstream loft-sh chart (#31). capi still has no CI
+  # infrastructure provider, and k0s has a harness pool but no smoke recipe —
+  # both tracked as follow-ups before they join the matrix;
+  # `backend_conformance_coverage_is_classified` (crd/profile.rs)
   # forces every new BackendType to be consciously classified against this
   # matrix. Since #28, reaching Ready on k3s/k0s/vcluster requires the
   # functional admission gates (DNSHealthy + InClusterToken), so every leg
@@ -201,7 +203,7 @@ jobs:
       matrix:
         # k3s-only on push (fast feedback on the production backend); the full
         # matrix nightly / on manual dispatch.
-        backend: ${{ fromJSON(github.event_name == 'push' && '["k3s"]' || '["k3s", "vkobe-etcd", "vkobe-kine"]') }}
+        backend: ${{ fromJSON(github.event_name == 'push' && '["k3s"]' || '["k3s", "vkobe-etcd", "vkobe-kine", "vcluster"]') }}
     steps:
       - uses: actions/checkout@v4
 
@@ -239,6 +241,7 @@ jobs:
             k3s)        mise exec -- just test-smoke-k3s "$CLUSTER" ;;
             vkobe-etcd) mise exec -- just test-smoke-vkobe ;;
             vkobe-kine) mise exec -- just test-smoke-vkobe-kine ;;
+            vcluster)   mise exec -- just test-smoke-vcluster ;;
             *) echo "unknown backend ${{ matrix.backend }}" >&2; exit 1 ;;
           esac
 
--- a/hack/e2e.ts
+++ b/hack/e2e.ts
@@ -801,6 +801,11 @@ spec:
     # cap, so a shorter value here expires the queued claim regardless of
     # how long the client waits (scale-to-zero pools provision on demand).
     queueTimeout: "12m"
+    # One provisioning attempt must fit inside this, or the operator
+    # recycles the instance as wedged before the claim can be served —
+    # capping the effective budget below the waits above. Ordering:
+    # creatingTimeout < LEASE_WAIT_TIMEOUT (recipe) < queueTimeout.
+    creatingTimeout: "8m"
   resources:
     limits:
       cpu: "500m"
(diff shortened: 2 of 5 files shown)


emmanuelm41:
Cross-model review (independent Codex and Claude legs, blind, same frozen diff) — **both independently found the same blocking defect**.

### The 10m/15m lease waits were unreachable

`scaling.queueTimeout` is a **server-side** cap: the lease controller expires any Pending claim older than it, regardless of the client's `--wait-timeout`. Every pool these recipes drive declared `queueTimeout: "5m"`:

| pool | queueTimeout (was) | recipe waits |
|---|---|---|
| `e2e-vkobe-etcd` | 5m | 10m |
| `e2e-vkobe-kine-sqlite` | 5m | 10m |
| `e2e-vcluster` | 5m | 15m |

So the real cold-provision budget was `min(wait, 5m)` = **5 minutes** on every leg, and the justfile comment claiming the vcluster wait was "correspondingly generous" was simply not true. A cold vcluster provision in kind — helm repo update, chart install, image pulls, guest apiserver, then the DNSHealthy + InClusterToken gates on a 30s health-check interval — is exactly the thing that overruns 5 minutes. The leg would have kept failing while appearing to have been given room.

That would have been a particularly annoying way to fail, too: it looks like the unwedge landed, but the leg sits `continue-on-error`-red for another cycle for a reason nobody would look for.

Server budget now exceeds the client wait on each driven pool (12m / 12m / 20m), so the client's timeout fires first with its clean message. The bootstrap pools already used 30m — the same reasoning, applied earlier.

### The likely failure also bypassed the diagnostics

A server-expired claim surfaces from the CLI as `"ended in phase Expired"`, not `"Timed out waiting for lease"` — so it missed the `printLeaseWaitDiagnostics` branch and rethrew bare, losing the pool/lease dump precisely in the "why didn't the pool provision?" case those diagnostics exist for. Now routed, with a message naming `queueTimeout` as the likely cause.

### Confirmed sound

Both reviews independently verified the tightened `backend_conformance_coverage_is_classified` mutations against the real `ci.yml`: exactly one line matches the matrix pattern, only the four `case` arms match the dispatch pattern, and each mutation fails the corresponding assertion where the old whole-file `ci.contains(leg)` caught neither.

### Noted, not changed

The arm check asserts an arm *exists*, not that it dispatches the *matching* recipe — `vcluster) … test-smoke-vkobe` would still pass. Tightening that means parsing the arm body, which felt like more machinery than the risk warrants; flagging it so the guard isn't over-trusted.

Rebased onto the corrected #57. 665 tests green.

Still not flipping `continue-on-error` here, for the reason in the description — and now with the added one that these legs have never actually completed a cold-start smoke, so a dispatch run is the thing that should decide it.

emmanuelm41:
Another round of cross-model review found the timeout chain was **still** incoherent — one layer further down than the last fix.

### `creatingTimeout` capped everything above it

The driven pools never set `scaling.creatingTimeout`, so it defaulted to **10m**. That is the cap on a *single provisioning attempt*: past it the operator recycles the instance as wedged, regardless of how long the claim may queue or the client is willing to wait.

So the vcluster leg's advertised 15m cold-start budget was really 10m per attempt, and the previous commit's 20m `queueTimeout` bought nothing. Exactly the same class of bug that commit fixed — three timeouts govern this path and I had only reconciled two of them.

The ordering that has to hold:

```
creatingTimeout  <  LEASE_WAIT_TIMEOUT (recipe)  <  queueTimeout
```

so one attempt can finish before the client gives up, and the client gives up before the server expires the claim — which also means the failure surfaces as the clean client-side timeout rather than a terminal `Expired` phase.

| pool | creatingTimeout | lease wait | queueTimeout |
|---|---|---|---|
| `e2e-vkobe-etcd` | 8m | 10m | 12m |
| `e2e-vkobe-kine-sqlite` | 8m | 10m | 12m |
| `e2e-vcluster` | 12m | 15m | 20m |

vcluster gets the larger attempt budget because its backend shells out to `helm --wait --timeout 300s` before the readiness gates even begin.

The relationship is now written next to each value. It's the kind of invariant that is invisible at any single site — each timeout looks reasonable alone, and only the composition is wrong — so it seemed worth stating rather than leaving to be re-derived.

### Status

Rebased onto `main` (which now carries #57 and #61), retargeted from the stacked base, and CI is green. Still not flipping `continue-on-error` on the non-k3s legs — that should follow a dispatch run that proves they actually complete a cold-start smoke, which is the thing none of this can establish from a unit test.

PR #60, opened 2026-08-07
fix(pool,lease): three correctness bugs found by audit

emmanuelm41:
Three independent defects found while auditing the pool/lease logic. Each has a test that was **confirmed red before the fix and green after**.

---

## 1. A deleted `Pending` lease deadlocks the entire pool

**`src/controllers/lease.rs`** — most severe of the three.

Every `remove_from_queue` call site needs a *live* lease to reconcile. A hard `DELETE` reaches none of them:

```rust
Err(kube::Error::Api(ae)) if ae.code == 404 => {
    debug!(lease = %name, "Lease disappeared before reconcile could load current state");
    return Ok(Action::await_change());   // <-- no queue eviction
}
```

The queue sorts `priority DESC, created_at ASC`, so the stranded entry sits at the **head**. Every later lease for that pool then computes `is_head == false` and returns before reaching `reserve_ready_instance`:

```rust
let head = queue.first().map(|h| h.lease_name == name).unwrap_or(false);
...
if !is_head { return Ok(Action::requeue(5s)); }
```

**No lease for that pool can bind again until the operator process restarts** — `rebuild_queues` runs only at startup, and only inserts.

Reachable via `kubectl delete clusterlease`, owner-reference GC, or kobe's own code: `create_lease` hard-deletes still-`Pending` leases on the quota-race and alias-race loser paths. (The normal client path is safe — `DELETE /v1/leases/{id}` patches `phase: Released` rather than deleting.)

**Fix:** evict from the queue on the 404 branch.

*Residual, not addressed here:* this relies on a reconcile actually being dispatched for the deleted object, which is the normal path. A periodic queue reconciliation would close the gap fully; that's a larger change and didn't seem warranted yet.

---

## 2. `backoff_delay_for` can return a **negative** delay

**`src/pool/manager.rs`**

```rust
let factor = (multiplier as u64).saturating_pow(shift);
let secs = base.num_seconds()
    .saturating_mul(factor as i64)      // u64::MAX as i64 == -1
    .min(max.num_seconds());
```

`factor` is deliberately saturating, so a large multiplier makes it `u64::MAX` — and `u64::MAX as i64` is `-1`. The delay flips sign. `.min(max)` doesn't catch it: it only bounds from *above*.

`next_attempt_at = now + delay` then lands in the **past**, `backoff_active` returns false, and a pool that has failed ~30 times consecutively resumes creating at full rate — precisely the condition the backoff exists to throttle.

The shipped default (`multiplier: 2`) is safe, which is why no existing test caught it. At `multiplier: 5` the delay goes negative from 29 consecutive failures onward:

```
multiplier=5 failures=29 produced a negative backoff of TimeDelta { secs: -5 }
```

**Fix:** `i64::try_from(factor).unwrap_or(i64::MAX)`. The new test is a property test over multipliers `[2, 3, 5, 10, 100, u32::MAX]` × failures `1..=64`, asserting the delay stays within `[0, cap]`.

---

## 3. Unstamped legacy members deadlock the rolling recycle

**`src/pool/manager.rs`**

Two predicates decide "is this drifted", and they disagree about `spec_hash == None`:

- `entry_drift_eligible` — an unstamped member past `UNSTAMPED_RECYCLE_GRACE` **is** drift-eligible
- `count_states` — partitions purely on hash equality, so unstamped counts as **clean**

The consequence is a permanent stall. Step 4 finds the members recycle-eligible, the `min_ready_during_upgrade` floor blocks the Delete, and the surge that is supposed to buy that headroom never fires — it's sized off `counts.ready_drifted`, which is `0`. The pool emits **no actions at all**, indefinitely:

```
unstamped legacy members are recycle-eligible but blocked by the floor,
so the pool must surge to buy headroom — got []
```

This silently breaks the promise in `entry_drift_eligible`'s own doc-comment — that legacy members get cleaned up "so future upgrades clean it up automatically instead of requiring a manual delete".

Affects kobe < 0.12.2 legacy members, and current-version members whose post-create status patch failed (after an operator restart, `build_pool_state` reads `None` from disk permanently). **PKI-expiring members hit the identical trap**, since `count_states` ignores cert expiry too.

**Fix:** size the surge off the step-4 candidate list, which already uses the correct predicate. Takes `.max()` of the two so every hash-drift path stays byte-identical — confirmed by the existing suite.

---

## Also found, not fixed here

A fourth issue was identified but is more invasive and lower confidence, so it's being filed separately: `sync_cluster_instance_statuses` writes `phase` from the stale in-memory entry while reading `lease_ref` fresh, so a cluster leased mid-reconcile can be written back to `Ready` with its `leaseRef` still set. `reserve_ready_instance` already defends the double-lease symptom (and its comment names this exact race), but `compute_pool_actions` has no lease awareness, so drift-recycle and idle scale-down can both target a cluster a tenant is holding.

## Verification

`cargo test --workspace` green (661), `cargo fmt -- --check` and `cargo clippy --workspace --all-targets` clean.

### DIFF
--- a/src/controllers/lease.rs
+++ b/src/controllers/lease.rs
@@ -216,14 +216,17 @@ async fn reconcile_lease<B: ClusterBackend + Clone + 'static>(
             Ok(current) => Arc::new(current),
             Err(kube::Error::Api(ae)) if ae.code == 404 => {
                 debug!(lease = %name, "Lease disappeared before reconcile could load current state");
-                // Evict it from the priority queue on the way out. Every
-                // other `remove_from_queue` call site needs a live lease to
-                // reconcile, so a hard DELETE (kubectl, owner GC, or the
-                // loser of a create race) would otherwise strand the entry
-                // permanently. Because the queue sorts oldest-first within a
-                // priority, that ghost sits at the head and every later lease
-                // for the pool sees `is_head == false` — so the pool stops
-                // binding entirely until the operator restarts.
+                // Evict on the way out. This covers only the narrow race
+                // where the delete lands after this reconcile was
+                // dispatched (store hit) but before its apiserver GET.
+                // It is NOT the fix for deletes in general: kube-runtime
+                // drives the controller from `applied_objects()`, which
+                // drops Deleted events, and requeues resolve through the
+                // reflector store — so a `kubectl delete` of a queued
+                // lease produces no reconcile here at all. The reaper's
+                // sweep (`prune_queues_against_live`) is what actually
+                // guarantees the entry goes away; this just gets there
+                // sooner when the race does happen.
                 remove_from_queue(&ctx.queues, &lease.spec.pool_ref, &name).await;
                 return Ok(Action::await_change());
             }
@@ -908,6 +911,36 @@ async fn run_reaper<B: ClusterBackend>(
 
         let now = chrono::Utc::now();
 
+        // Reconcile the in-memory queues against what the apiserver
+        // actually holds. This is the only place a lease deleted outside
+        // a reconcile can be evicted — see `prune_queues_against_live`
+        // for why the reconcile path cannot do it — and a stranded entry
+        // head-blocks its whole pool, so it is worth the one pass over an
+        // already-fetched list.
+        {
+            let live_pending: std::collections::HashSet<String> = leases
+                .iter()
+                .filter(|l| {
+                    l.status
+                        .as_ref()
+                        .map(|s| s.phase == LeasePhase::Pending)
+                        .unwrap_or(true)
+                })
+                .map(|l| l.name_any())
+                .collect();
+
+            let evicted = {
+                let mut queues = ctx.queues.write().await;
+                prune_queues_against_live(&mut queues, &live_pending)
+            };
+            if !evicted.is_empty() {
+                warn!(
+                    leases = ?evicted,
+                    "Reaper: evicted queue entries with no live Pending lease (would otherwise head-block the pool)"
+                );
+            }
+        }
+
         for lease in leases {
             let name = lease.name_any();
             let status = lease.status.clone().unwrap_or_default();
@@ -973,6 +1006,45 @@ async fn remove_from_queue(
     if let Some(queue) = queues.get_mut(profile) {
         queue.retain(|p| p.lease_name != lease_name);
     }
+}
+
+/// Drop queue entries whose lease is no longer `Pending` on the
+/// apiserver, returning the names evicted.
+///
+/// This is the authoritative cleanup, and it exists because the
+/// reconcile path cannot be. kube-runtime's `Controller` is driven by
+/// `applied_objects()`, which drops Deleted events, and every scheduled
+/// requeue resolves through the reflector store first — so once a
+/// deleted lease leaves the store, **no reconcile ever runs for it**.
+/// The 404 branch in `reconcile_lease` therefore only catches the narrow
+/// race where a delete lands mid-reconcile; a plain `kubectl delete` of
+/// a queued lease never reaches it.
+///
+/// That matters because a stranded entry is not a leak but a deadlock:
+/// the queue is sorted oldest-first within a priority, so the ghost sits
+/// at the head and every later lease for that pool sees `is_head ==
+/// false` and never binds.
+///
+/// Sweeping from a LIST is safe against the obvious race — a lease
+/// created after the LIST could be pruned here, but the queue insert in
+/// `reconcile_lease` is idempotent and runs on a 5s requeue, so it
+/// re-inserts itself on the next pass. A transient drop self-heals; a
+/// permanent ghost does not.
+fn prune_queues_against_live(
+    queues: &mut HashMap<String, Vec<PendingLease>>,
+    live_pending: &std::collections::HashSet<String>,
+) -> Vec<String> {
+    let mut evicted = Vec::new();
+    for queue in queues.values_mut() {
+        queue.retain(|p| {
+            let keep = live_pending.contains(&p.lease_name);
+            if !keep {
+                evicted.push(p.lease_name.clone());
+            }
+            keep
+        });
+    }
+    evicted
 }
 
 /// Build the `{ "status": { ... } }` merge patch that transitions a lease to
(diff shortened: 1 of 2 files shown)


emmanuelm41:
Cross-model review (independent Codex and Claude legs, generated blind from the same frozen diff) — **both independently caught the same factual error**, and they were right.

### The cert-expiry claim was false

This PR claimed `count_states` ignores PKI expiry and that cert-expiring members hit the same deadlock as unstamped ones. Checked against the base commit:

```
$ git show ab09d99:src/pool/manager.rs | rg 'let drifted = '
1093:            let drifted = hash_drifted || entry.cert_expiring();
```

So cert-expiring members were **always** counted in `ready_drifted` and always got their surge — and `cert_expiring_ready_recycles_like_drift_leased_untouched` already pins exactly that. There was no PKI trap to fix.

Worse, I'd shipped that falsehood as two code comments sitting 400 lines from the implementation that contradicts them, at the most subtle decision point in the pool manager. `AGENTS.md` puts it plainly — a doc-comment that contradicts the body is worse than no doc-comment. Both rewritten.

### The `.max()` was a no-op dressed as a safety measure

One review showed it's provable rather than merely usually-true: `drifted_ready` is every non-`deleting` Ready member that is hash-drifted, cert-expiring, or unstamped-past-grace, and nothing Ready can be in `deleting` at that point. So it's a strict superset of `ready_drifted` and the max could never pick the other operand. Replaced with the plain value, and the reasoning is now written down instead of the hand-wavy "taking the larger of the two".

### An undisclosed side effect, now stated

`drift_in_flight` also feeds `upgrade_in_progress`, so a pool holding unstamped members now suppresses idle scale-down until they recycle. That's intended — scale-down would otherwise reap the surge replacements this fix buys — but it's a real change for pools carrying legacy members and the description didn't mention it.

### What held up

Both reviews independently confirmed all three bugs are real and the fixes correct, and one reproduced the red-before/green-after claim by reverting the fix hunks and watching the tests fail with the exact messages. The backoff fix was also noted to cover a case I hadn't mentioned: `multiplier=u32::MAX, failures=3` overflows `i64` without saturating `u64`, and `try_from` handles that too.

No behavior change in this follow-up commit — 661 tests still green.

emmanuelm41:
Round 2 of the cross-model review found that the ghost-lease fix **did not do what its comment claimed**, and this one is worth spelling out because it's the most consequential finding of the batch.

### The 404 branch almost never runs

I verified this against kube-runtime 3.1.0 as resolved in the lockfile, not from memory:

- `Controller::new` drives from `reflector(writer, watcher(..)).applied_objects()`, and `applied_objects` is `EventDecode::new(self, false)`. The `false` is the thing: the sibling `touched_objects` passes `true` and is documented as *"All Added/Modified/Deleted events are passed through"*. So **Deleted events are dropped**.
- Every scheduled requeue resolves `store.get(&request.obj_ref)` first and returns `Error::ObjectNotFound` without calling the reconciler.

So after a plain `kubectl delete` of a queued lease, the watch removes it from the store and **no reconcile with a 404 ever happens**. The branch I added only caught the race where a delete lands after a reconcile was dispatched but before its apiserver GET.

Which means the headline case — the pool deadlock this PR exists to fix — was still broken, while the comment asserted it was handled. Same failure mode as the cert-expiry claim earlier in this PR: a confident comment that the code doesn't support. My `disappeared_pending_lease_is_evicted_from_the_queue` test couldn't catch it either, because it calls `reconcile_lease` directly and so never exercises the dispatch path that doesn't route there.

### The fix now lives in the reaper

The reaper already LISTs every lease each 60s, so it can reconcile the in-memory queues against the live set. `prune_queues_against_live` drops any queued name that isn't a live `Pending` lease — across **every** pool queue, not just the deleted lease's own pool, which is all the reconcile path could ever have known about.

Race-safe in the direction that matters: a lease created after the LIST could be pruned, but the queue insert in `reconcile_lease` is idempotent and runs on a 5s requeue, so it re-inserts itself next pass. A transient drop self-heals; a permanent ghost doesn't.

The 404 eviction stays as a fast path for when the race does happen, with its comment corrected to claim only that.

### Also in this round

`count_states` folding `cert_expiring()` was confirmed at the base commit, so the corrected comments are accurate. The `.max()` removal was independently confirmed sound (nothing Ready can be in `deleting` before step 4, so `drifted_ready` is a strict superset). And the backoff fix was noted to cover a case beyond the one described: `multiplier=8, failures=22` gives `8^21 = 2^63`, which overflows `i64` *without* saturating `u64` — `try_from` handles that too.

663 tests green, fmt and clippy clean.

emmanuelm41:
Round 3 of the cross-model review: **fable says SHIP**, codex says block on one point, which I checked and believe does not hold.

### The contested point

codex's blocker: *"name-only queue pruning leaves a permanent ghost after delete/recreate of the same lease name."* The concern is that `prune_queues_against_live` matches on lease name, so if a queued lease `X` is deleted and a **new** lease also named `X` is created, the stale entry survives the sweep carrying the old `created_at`.

That's mechanically true, but it isn't reachable in practice. Lease names are server-generated:

```rust
let lease_id = format!("lease-{}", &uuid::Uuid::new_v4().to_string().replace('-', "")[..12]);
```

48 bits of random, not caller-supplied — so "delete a lease and recreate one with the same name" isn't an operational scenario. And if it somehow happened, the consequence would be a stale queue *position* (the new lease inherits the old `created_at`, so it sorts earlier than it should) — a fairness wrinkle, not the head-blocking deadlock this PR fixes.

I'd rather record the reasoning than silently disagree: if lease naming ever becomes caller-controllable, this becomes real, and the fix would be to key queue entries on UID rather than name.

### What round 3 confirmed

fable independently verified the parts that matter most, and found them sound:

- All three bugs are real, and each fix correct and complete.
- The reaper sweep is **race-safe** on every interleaving it checked: namespaces align across the watch, `rebuild_queues` and `run_reaper`; a lease created after the LIST is re-inserted within ≤5s by the idempotent insert, preserving its original `creation_timestamp`; statusless leases are conservatively kept; a delete landing after the LIST is caught on the next sweep (≤~120s).
- Every shipped comment is now factually true of the code — including the corrected cert-expiry claim and the re-scoped 404-branch comment.
- It also noted the backoff fix covers a case my description understates: the hazard is any factor above `i64::MAX`, not only the saturated `u64::MAX`, and `multiplier = u32::MAX` triggers the *unsaturated* overflow at just 3 failures. The fix and the property test cover the whole class.

### One follow-up worth doing, not blocking

The `live_pending` filter is built inline in `run_reaper`, so it's untested — `prune_queues_against_live` is tested pure, but an inverted filter in the loop (which would evict *live* Pending leases) wouldn't be caught. Extracting that filter into a testable function is the obvious next step; a 60s-interval loop is awkward to unit-test as-is.

Leaving the merge decision open rather than acting on a split verdict.

PR #61, opened 2026-08-07
test: cover the CRD instance types, CLI config, and kobe-sync syncers

emmanuelm41:
Adds **88 tests** across the five files with zero or near-zero coverage relative to their size. Companion to #58 (which covered `src/backend/vcluster.rs`); no overlap.

| File | Before | After |
|---|---|---|
| `src/crd/instance.rs` | **0** | 22 |
| `crates/kobectl/src/commands/config.rs` | 2 | 32 |
| `src/kobe_sync/syncer/pvcs.rs` | 4 | 19 |
| `src/kobe_sync/syncer/service_accounts.rs` | 4 | 17 |
| `src/kobe_sync/bootstrap.rs` | 3 | 11 |

## What's pinned, and why it matters

**`src/crd/instance.rs`** — mostly serde contracts that are load-bearing and fail *silently*:

- The Merge-Patch split between fields that must serialize `None` as an explicit `null` to clear them (`leaseRef`, `idleSince`, `stateSince`) and fields that must be **omitted** to survive a patch (`specHash`, `createdWith`, `message`, `activeBootstrap`, `network`). Getting this backwards either fails to release a lease or erases provenance.
- A condition's Rust `condition_type` serializes as `type` — that's what `kubectl wait --for=condition=` reads.
- Empty `conditions` is omitted, so a foreign status writer can't array-replace the list away.
- Forward and backward compat: a status written by a *newer* operator still deserializes (rolling-upgrade safety), and so do statuses from kobe < 0.17 and provenance from < 0.23.1.
- The generated CRD keeps its identity (group / kind / plural / `ci` shortname / single served+storage version / status subresource) and types `specHash` as a **string**, per the float64 structural-validator constraint its doc-comment explains.

**`crates/kobectl/src/commands/config.rs`** — resolution precedence (`--target` > session > legacy `current_target` > flat legacy `endpoint`), the local/global overlay and its `Scope` reporting, and legacy migration being a no-op in each case where it could otherwise destroy a token. Also that `--endpoint` keeps the target's auth rather than silently downgrading to anonymous, and that token masking can never reveal a usable secret.

**`pvcs.rs` / `service_accounts.rs`** — translation purity, plus each syncer's event handling against a wiremock apiserver: create when absent, forced SSA when present, **skipped namespaces producing zero host traffic on any event kind**, delete tolerating 404 but propagating other errors, and untranslatable names never reaching the host.

**`bootstrap.rs`** — the RBAC kobe-sync grants itself: no wildcard verbs/resources, no write verbs on read-only watch targets, both objects applied as forced SSA with the role before the binding. This runs with `system:masters`, so the shape of that Role is the only thing bounding it.

## Mutation-checked

Nine seeded regressions, all caught, all reverted:

| Mutation | Caught by |
|---|---|
| add `skip_serializing_if` to `lease_ref` | `actively_managed_fields_serialize_none_as_explicit_null` |
| remove `skip_serializing_if` from `spec_hash` | `write_once_fields_are_omitted_when_none_…` |
| `shortname = "ci"` → `"cinst"` | `crd_identity_is_stable` |
| drop `.force()` from `PatchParams::apply` | `ensure_rbac_uses_forced_server_side_apply_as_kobe_sync` |
| SA delete `err.code == 404` → `410` | `delete_event_tolerates_a_host_sa_that_is_already_gone` |
| disable the SA skip-namespace guard | `skipped_namespaces_produce_no_host_traffic_at_all` |
| PVC `status: None` → `pvc.status.clone()` | `translate_pvc_drops_a_populated_virtual_status` |
| mask boundary `len() > 8` → `>= 8` | `token_display_never_reveals_a_usable_secret` |
| overlay `Scope::Both` → `Scope::Local` | `overlay_local_target_shadows_…_marked_both` |

## Production changes — two, both minimal

1. **`mask_token` extracted** out of `print_auth` (config.rs). Logic byte-identical; it was unreachable from a test because `print_auth` only writes to stdout, and a credential-leak guard is worth pinning.
2. **`mock_k8s_client` passes an empty `root_cert`** (`src/testutil.rs`, `#[cfg(test)]`-gated). This fixes a **pre-existing flake**, not something this PR introduced: without it kube falls through to rustls' `with_native_roots()`, and under a heavily parallel `cargo test` macOS' Security framework intermittently fails the trust-settings read with `Os(-36) I/O error`, surfacing as `RustlsTls(NoValidNativeRootCA)`. Reproduced twice locally — three `backend::k3s` tests failed under load and passed 3/3 in isolation. The mock server speaks plain HTTP, so the trust store is never consulted. **No production TLS path was touched.** Worth knowing if CI ever moves to macOS runners.

`kobe-sync` can't reuse `src/testutil.rs` (`kobe_sync_bin` deliberately doesn't pull in the operator's module tree), so it gets its own `#[cfg(test)] mod testkit`.

## Documented limitation pinned, not changed

`config.rs` `overlay` applies a local `auth` only when it differs from the default, so a `.kobe.toml` that explicitly writes `auth = "oidc"` can't pull a global `token` target back to OIDC. TOML can't distinguish "absent" from "explicitly default", so this is arguably intentional — `overlay_cannot_reset_auth_to_the_default_mode` pins it so it stays a decision rather than drifting.

## Verification

`cargo test --workspace` green (1112 total), `cargo fmt -- --check` and `cargo clippy --workspace --all-targets` clean.

### DIFF
--- a/crates/kobectl/src/commands/config.rs
+++ b/crates/kobectl/src/commands/config.rs
@@ -903,18 +903,25 @@ fn print_config(config: &CliConfig, target_override: Option<&str>) -> Result<()>
     Ok(())
 }
 
+/// Render a bearer token for display, never revealing enough of it to
+/// be usable.
+///
+/// Tokens longer than 8 characters keep a 4-char head and tail so a user
+/// can tell *which* token is configured; anything shorter is replaced
+/// wholesale, because a 4+4 elision of an 8-char secret would print the
+/// entire value. `None` renders as `(not set)`.
+fn mask_token(token: Option<&str>) -> String {
+    match token {
+        Some(t) if t.len() > 8 => format!("{}...{}", &t[..4], &t[t.len() - 4..]),
+        Some(_) => "****".to_string(),
+        None => "(not set)".to_string(),
+    }
+}
+
 fn print_auth(auth: &AuthMode, token: Option<&str>, ssh_fingerprint: Option<&str>) {
     println!("auth:     {auth}");
     if auth == &AuthMode::Token {
-        let masked = token
-            .map(|t| {
-                if t.len() > 8 {
-                    format!("{}...{}", &t[..4], &t[t.len() - 4..])
-                } else {
-                    "****".to_string()
-                }
-            })
-            .unwrap_or_else(|| "(not set)".to_string());
+        let masked = mask_token(token);
         println!("token:    {masked}");
     }
     if auth == &AuthMode::Ssh {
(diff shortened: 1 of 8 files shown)


emmanuelm41:
Cross-model review found a real defect this PR introduced, which the full-suite green run was hiding.

### The new kobe-sync tests were order-dependent

`testkit::mock_client` built a `kube::Client` without installing a process-level `CryptoProvider`. This build enables both `ring` and `aws-lc-rs`, so rustls can't choose one and `Client::try_from` panics.

The suite passed only because some earlier test in the same binary happened to install a provider first. Running one on its own:

```
$ cargo test --bin kobe-sync -- --exact kobe_sync::bootstrap::tests::ensure_rbac_uses_forced_server_side_apply_as_kobe_sync
panicked at rustls-0.23.37/src/crypto/mod.rs:249:
Could not automatically determine the process-level CryptoProvider from Rustls crate features.
```

That's the normal inner-loop command during development, so it would have been a confusing first encounter for anyone touching these files.

The operator's own helpers already install it (`controllers/lease.rs`, `pki.rs`) — the new kobe-sync helper was the only one missing it, so this was introduced here rather than pre-existing. I verified the operator side passes in isolation, confirming the asymmetry.

Fixed, and re-verified: the same isolated test now passes, and the full workspace suite stays green.

### Also raised, not changed

- `token_display_never_reveals_a_usable_secret` overclaims: a 9-character token reveals 8 of 9 characters. The masking behavior is inherited, not introduced here, but the test name promises more than it delivers — worth renaming.
- `resolve_prefers_the_session_target_over_the_config_field` can return without asserting if session saving fails in some environments.

Both are test-naming/robustness nits rather than correctness problems, so I've left them for a follow-up rather than expanding this PR further.

Issue #53, opened 2026-07-27
chore(chart): list the kobe Helm chart on Artifact Hub


0xPxt:
Now that the kobe Helm chart is published as an OCI artifact to Docker Hub (`oci://registry-1.docker.io/zondax/kobe`, versioned with the operator — currently `0.37.0`), we can also list it on [Artifact Hub](https://artifacthub.io) for discoverability. Artifact Hub **indexes** the chart from the OCI registry; it does not host it.

**Prerequisite:** the `zondax/kobe` Docker Hub repo must be **public**, and chart tags must be valid semver (`0.37.0` qualifies).

**Steps** (verified against the Artifact Hub docs):

1. Create / sign in to an Artifact Hub account and a `zondax` organization.
2. Add a repository → kind **Helm**, URL `oci://registry-1.docker.io/zondax/kobe`. Artifact Hub indexes each semver-tagged version.
3. Claim ownership + earn the **Verified Publisher** badge: take the `repositoryID` Artifact Hub generates for the repo, put it in an `artifacthub-repo.yml`, and push it to the OCI repo as a metadata artifact with `oras`:
   ```
   oras push registry-1.docker.io/zondax/kobe:artifacthub.io \
     --config /dev/null:application/vnd.cncf.artifacthub.config.v1+yaml \
     artifacthub-repo.yml:application/vnd.cncf.artifacthub.repository-metadata.layer.v1.yaml
   ```
4. (Optional) enrich `charts/kobe/Chart.yaml` with `artifacthub.io/*` annotations (changelog, license, maintainers, links, images, security) for a richer listing.

**Notes:**
- Steps 1–2 are account/UI actions; 3–4 are small repo changes plus a one-time `oras push`.
- OCI-hosted Helm repos on Artifact Hub can't force-reindex an existing version by digest change — a limitation, not a blocker.

Ref: https://artifacthub.io/docs/topics/repositories/helm-charts/

emmanuelm41:
Did the in-repo half in #64: `artifacthub.io/*` annotations on `Chart.yaml` — license, operator, prerelease, links, the full image list for the security report, and all eight CRDs (cross-checked against `charts/kobe/crds/`).

Prerequisites all still hold: `zondax/kobe` on Docker Hub is public, `0.37.0` is valid semver, and the chart isn't listed on Artifact Hub yet.

[...]

Issue #17, opened 2026-07-17
docs: reconcile runtime strategy — kobe-sync retired in favor of vcluster-backend + native pkobe


jleni:
## Context

The runtime strategy has moved to **vcluster-as-a-backend (density) + native pkobe (isolation)**, with the in-house kobe-sync syncer retired. The public docs don't reflect this yet — they still present kobe-sync as a core component:

- `docs/kobe-docs/how-it-works/architecture.mdx` lists kobe-sync as one of the four main components and documents it as the per-cluster sidecar/runtime.
- `docs/kobe-docs/pools/backends.mdx` describes the vkobe backend as built on kobe-sync.

Contributors currently find guidance that treats the retired path as current.

## Proposal

- Add/update a single authoritative runtime-strategy doc: vcluster = density backend; pkobe = native isolation backend; kobe-sync = retired.
- Update `architecture.mdx` and `backends.mdx` accordingly.
- Finish/confirm the vkobe→vcluster backend migration so there is one density backend.

## Acceptance

One authoritative strategy doc; no live doc treating kobe-sync as the path.

PR #64, opened 2026-08-07
chore(chart): add Artifact Hub annotations and gate their image tags

emmanuelm41:
The in-repo half of #53. Prerequisites are already met — `zondax/kobe` on Docker Hub is public, holds a valid semver tag (`0.37.0`), and the chart isn't listed on Artifact Hub yet.

## What's here

`charts/kobe/Chart.yaml` gains:

- `artifacthub.io/license`, `artifacthub.io/operator`, `artifacthub.io/prerelease`
- `artifacthub.io/links` — source, docs, issues
- `artifacthub.io/images` — every image the chart can deploy, so Artifact Hub's security report scans the right things
- `artifacthub.io/crds` — all eight CRDs, with display names and one-line descriptions

The CRD list was cross-checked against `charts/kobe/crds/`: kinds, plural names and versions all match exactly.

## The drift hazard, and the gate for it

`artifacthub.io/images` pins fully-qualified refs, so the first-party ones carry the release version:

```yaml
- name: kobe-operator
  image: docker.io/zondax/kobe-operator:0.37.0
```

Left alone, that would silently advertise a stale tag forever after the next bump — the worst kind of metadata bug, since the listing keeps looking correct. So this PR also:

- teaches `just bump` to rewrite those tags alongside `version` / `appVersion`
- extends `scripts/check-version-consistency.sh` to gate them

The check fails on a desynced tag:

```
version consistency FAILED for the workspace manifests:
  - Chart.yaml artifacthub.io/images kobe-sync tag '0.36.0' != workspace version '0.37.0'
```

It also fails if the annotation is removed entirely, so dropping it has to be a conscious act rather than a silent loss of coverage. Third-party images (`kine`) are deliberately not version-gated — they don't move with our releases. That gate already runs in CI via the `version-consistency` job, which `publish-chart` depends on.

## Description and keywords refreshed

The description read:

> Supports multiple backends: k3s, k0s, CAPI, and kobe-sync.

kobe-sync is the sidecar the vkobe backend uses, not a backend type, and it's retired as a strategy (#17) — while **vcluster**, the actual direction of travel, was missing entirely. Same fix in the keywords. This text is the chart's front page on Artifact Hub, so it seemed worth getting right before anyone reads it there.

## What still needs a human

Steps 1–2 of #53 are account/UI actions and can't be done from the repo:

1. Create/sign in to Artifact Hub, create the `zondax` org
2. Add the repo `oci://registry-1.docker.io/zondax/kobe` (kind: Helm)

Only then does Artifact Hub issue the `repositoryID`, which is the **input** to `artifacthub-repo.yml` — so step 3's file genuinely cannot be written correctly before that. The subsequent `oras push` also needs Docker Hub write creds (`DOCKERHUB_USER`/`DOCKERHUB_TOKEN` already exist as CI secrets).

So #53 should stay open after this merges, blocked on that registration.

## Verification

`helm lint charts/kobe` passes. All three multi-line annotations parse as YAML (3 images, 8 CRDs, 3 links). The version gate passes clean and was confirmed to fail on a deliberately desynced tag.

Refs #53

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -90,6 +90,13 @@ jobs:
       - name: Tests
         run: cargo test --workspace
 
+      # The release gate itself only runs on tag pushes (version-consistency),
+      # which is the right place for it — but that means its parsing logic is
+      # otherwise unexercised until a release is being cut. These tests run on
+      # every change so a regression in the gate surfaces here instead.
+      - name: Version-consistency gate tests
+        run: ./scripts/test-version-consistency.sh
+
       - name: Security audit
         run: |
           cargo install cargo-audit --locked || true
--- a/charts/kobe/Chart.yaml
+++ b/charts/kobe/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: kobe
 description: >-
   Kubernetes operator for instant cluster provisioning via warm pools.
-  Supports multiple backends: k3s, k0s, CAPI, and kobe-sync.
+  Supports multiple backends: k3s, k0s, vcluster, and CAPI.
 type: application
 version: 0.37.0
 appVersion: "0.37.0"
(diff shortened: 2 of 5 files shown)


emmanuelm41:
One scoping note on the gate, so it isn't over-read: `version-consistency` is tag-gated (`if: startsWith(github.ref, 'refs/tags/v')`), so the new image-tag check runs **at release time, not on PRs**. That's inherited from how the job already worked, not something this PR changes.

I think that's the right placement — it fires before anything irreversible (signing, GitHub Release, crates.io, chart push), and the realistic failure mode is "`just bump` ran but the annotation rewrite didn't", which is precisely a release-time concern. But it does mean a PR that hand-edits an annotation tag out of sync won't be caught until someone cuts a tag.

Verified locally instead: desyncing `kobe-sync` to `0.36.0` fails the check with a specific message, and restoring it passes.

emmanuelm41:
Cross-model review (independent Codex and Claude legs, blind, same frozen diff) — **both independently found the same blocking defect**, and I confirmed it against the live registry.

### The pinned image tags did not exist

`docker-bake.hcl` publishes `latest`, `dev`, `v${VERSION}`, `v0.37`, `v0` — never a bare `0.37.0`. The chart itself deploys `printf "v%s" .Chart.AppVersion`. Docker Hub agrees:

```
['latest', 'dev', 'v0', 'v0.37', 'v0.37.0', 'v0.35', ...]
```

So the annotation pointed Artifact Hub's scanner at refs that resolve to nothing — defeating the one thing it was added for. I'd conflated the **chart's** OCI tag (bare semver, which is what `helm push` creates, and what I cited as "valid semver" in the description) with the **container images'** tag scheme.

The tooling made it worse rather than catching it: the new gate compared against the bare version and `just bump` rewrote to the bare version, so both would have *rejected the correct tag* and restored the broken one. As one review put it, it didn't just miss the bug, it institutionalized it. My local verification passed only because the annotation and the gate were consistently wrong together.

All three now expect `v<version>`, and the failure message says why. Verified: passes on v-prefixed tags, fails on the bare tags this removes, and round-trips through a simulated bump to 0.38.0.

### "Every image this chart can deploy" was not true

The chart also deploys `ghcr.io/fluxcd/flux-cli:v2.2.3` (`bootstrap-flux.yaml`) and `rancher/kubectl:v1.34.7` (`job-psa-labels.yaml`), both missing while conditional `kine` was included. Both added, whitelisted as third-party.

### Confirmed sound

Both reviews independently verified the annotation syntax against Artifact Hub's documented Helm schema, and the 8-CRD cross-check against `charts/kobe/crds/`. One also independently reproduced all three gate behaviors (pass, desynced-tag failure, annotation-removed failure).

Two notes I'm leaving as-is: `artifacthub.io/prerelease` is static, so an rc release would carry `"false"` — worth fixing when rc chart publishing matters, not now. And the gate is tag-time only, which I noted above.

Issue #18, opened 2026-07-17
Pluggable multi-driver DataStore (PostgreSQL/MySQL/NATS) over KobeStore


jleni:
## Context

The `KobeStoreDriver` enum is currently only `Etcd` + `KineSqlite`. A decoupled, pluggable datastore (etcd/MySQL/PostgreSQL/NATS) with one store backing many control planes — the pattern Kamaji has proven out — is table-stakes for production HA across backends, and a natural fit for the existing `KobeStore` abstraction.

## Proposal

- Add PostgreSQL/MySQL/NATS drivers (kine-fronted) to `KobeStore`.
- Optional shared-datastore key-prefix multi-tenancy; enforce schema/username uniqueness (cf. clastix/kamaji#889) and pin against kine version drift (cf. clastix/kamaji#1101).

## Acceptance

A pool can run on an external PostgreSQL/MySQL datastore; driver selectable per `KobeStore`.

PR #87, opened 2026-08-10
test: cover the CRD types and CLI helpers that had no tests

emmanuelm41:
`access_policy.rs`, `cidr.rs` and `datastore.rs` had **zero tests** between them. All three encode contracts that fail *silently* when broken — the type still compiles, the manifest still applies, and the behaviour changes.

## `access_policy.rs` — 10 tests

This is the authorization CRD, so the failure modes are the interesting ones:

- **`match_clause` must serialize as `match`.** The rename isn't cosmetic — `match` is a Rust keyword, so the field *has* to be renamed, and that rename is the entire multi-role OIDC mechanism. Lose it and a rule scoped to one role deserializes with **no match clause**, which means it applies to every caller the policy authenticates. An authorization widening, with nothing else in the type looking wrong.
- **`maxTtl`, `maxConcurrentLeases`, `pools` are required on purpose.** A default on any of them turns "this policy forgot its ceiling" from a rejection into a silent value nobody chose.
- **`identity` defaults to `{sub}`** — it decides who a caller is taken to be.
- **OIDC `algorithms` defaults to RS256**, asserted non-empty and never an `HS*` family. An empty or symmetric default is the classic algorithm-confusion footgun.

## `cidr.rs` — 6 tests

- **Allocated CIDRs must be omitted when unset, never null.** The allocator derives its in-use set by listing bound claims, so a Merge-Patch that erases one doesn't just lose a display value — the slot reads as *free* and gets re-issued to a second cluster. That's the collision class the module's own doc-comment warns about.
- **Phase defaults to `Pending`**, so an unreconciled claim can't advertise an allocation it doesn't have.

## `datastore.rs` — 3 tests

- **Driver wire values are kebab-case explicit renames** that `rename_all = "camelCase"` would not produce for `KineSqlite`. Losing the rename makes every stored `kine-sqlite` KobeStore unreadable — and an unreadable datastore stops being reconciled while the guests depending on it keep running.
- **The supported set is pinned** so it can't widen by accident; adding a driver is deliberate work (#18).

## Mutation-checked

Five seeded regressions, each caught:

| mutation | caught by |
|---|---|
| drop the `match` rename | `match_clause_is_named_match_on_the_wire` |
| give `maxTtl` a default | `the_lease_ceilings_have_no_default_and_must_be_declared` |
| default algorithms to `HS256` | `oidc_algorithms_default_to_an_asymmetric_algorithm` |
| serialize an unset CIDR as null | `an_unset_allocation_is_omitted_so_merge_patch_cannot_erase_it` |
| drop the `kine-sqlite` rename | `driver_wire_values_are_the_kebab_case_names` |

## Verification

`cargo test --workspace` green (755 operator, up from 736), `cargo fmt --check` and `cargo clippy --workspace --all-targets` clean.

### DIFF
--- a/crates/kobectl/src/commands/status.rs
+++ b/crates/kobectl/src/commands/status.rs
@@ -259,3 +259,37 @@ async fn enrich_leases(
 
     enriched
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    /// The SSH hint fires on the one failure it can actually explain, and
+    /// stays quiet otherwise.
+    ///
+    /// "no key found in SSH agent" is the failure a user cannot diagnose from
+    /// the message alone — the fix is environmental (`SSH_AUTH_SOCK`,
+    /// `ssh-add -l`), not something they typed wrong. Attaching that hint to
+    /// unrelated auth failures would send people chasing their agent when the
+    /// real problem is an expired token or a bad issuer.
+    #[test]
+    fn the_ssh_hint_is_offered_only_for_the_failure_it_explains() {
+        assert!(
+            auth_error_hint("no matching key found in SSH agent").is_some(),
+            "the agent-lookup failure is the case this hint exists for"
+        );
+
+        for unrelated in [
+            "token expired",
+            "invalid issuer",
+            "connection refused",
+            "403 Forbidden",
+            "",
+        ] {
+            assert!(
+                auth_error_hint(unrelated).is_none(),
+                "must not offer an SSH-agent hint for: {unrelated:?}"
+            );
+        }
+    }
+}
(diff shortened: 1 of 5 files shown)


emmanuelm41:
Extended with two CLI helpers that were also untested — both small, both load-bearing:

**`with_lease::exit_code`** — this is the whole contract of the wrapper in CI. `kobe with-lease -- <cmd>` *stands in for* `<cmd>`, so anything that isn't a clean coded exit must map to non-zero. A signal-killed child has no exit code at all, and a failed wait means the outcome was never learned. Mapping either to `0` turns a crashed test suite into a green pipeline — the one failure mode a CI wrapper must not have. Clean codes pass through verbatim, since callers branch on the specific value.

Mutation-checked: flipping the fallback to `0` fails the signal case.

**`status::auth_error_hint`** — fires only for the SSH-agent lookup failure, which is the one a user can't diagnose from the message alone (the fix is environmental: `SSH_AUTH_SOCK`, `ssh-add -l`). Attaching it to an expired token or a bad issuer would send people to check their agent for a problem that isn't there.

Suite now 755 operator + 230 kobe-sync + 63 kobectl, all green.

Issue #103, opened 2026-08-11
lifecycle(k3s): surface quarantine and gate its exit on a verified receipt


emmanuelm41:
Found by cross-family review of PR #96 (#80 layer 1). PR #96 added `ClusterState::Quarantined`, excluded it from the pool's unconditional Unhealthy delete, and made it count against the capacity ceiling. That is containment only — quarantine is still invisible and has no way out.

## Quarantine is not observable

- [ ] Pool phase ignores it. A `minReady: 0` pool containing only quarantined instances reports `Idle`; a pool that otherwise meets `minReady` reports `Healthy`. Neither is an honest readiness signal.
- [ ] Pool status, API responses, CLI output, and the `kobe_pool_size` metric do not expose the state at all.
- [ ] The `StateCounts` doc comment still describes the original counters as summing to pool size, which is now false.

An operator whose teardown evidence is stuck should be able to see that, not infer it.

## "Never deleted" is narrower than it sounds

PR #96 only protects quarantine inside `compute_pool_actions`. Nothing else refuses:

- [ ] Direct deletion of a quarantined `ClusterInstance` or its `ClusterLease`.
- [ ] Finalizer removal without evidence. Comments assert that finalizers and binding provenance are retained; nothing enforces that a user, GC, or stale controller cannot delete the object and erase the quarantine record.
- [ ] Status writes are merge patches without a phase/UID precondition, so a stale reconcile can overwrite `Quarantined` with `Recycling`.

## There is no exit

- [ ] No transition consumes a valid receipt and releases the quarantine. The instance controller must own exactly one receipt-gated path out; a wildcard phase arm is not an enforcement mechanism.

## Receipt durability

- [ ] The receipt lives on `ClusterLeaseStatus`, chosen so it outlives the `ClusterInstance`. But the normal recycling flow deletes the **lease** once the instance 404s — so a verified receipt can disappear before #74's owning `SandboxLease` consumes it. Either add a lease-level finalizer held until acknowledgement, or move the receipt to a separate object retained until an explicit consumer ack.

## Mode immutability

- [ ] `cleanupMode` is mutable, tenant-visible spec. A lease bound as `VerifiedDestroy` must not be downgradeable to `Standard` afterwards. Enforce via admission/CEL plus API-layer policy, and bind the selected mode into immutable provenance.
[...]

Issue #102, opened 2026-08-11
lifecycle(k3s): bind-time creation manifest so receipts prove concrete resources


emmanuelm41:
Found by cross-family review of PR #96 (#80 layer 1). PR #96 tightened receipt validation so a receipt must carry one `Verified` check per subject in its scope. This issue supplies the thing that makes that meaningful.

## The gap

`permits_release_for` compares a receipt against a `TeardownScope`. Nothing populates a scope from real creation-time facts yet, so **nothing calls it in production**. More importantly, the validation only proves the receipt agrees with the scope — so a scope that is wrong, incomplete, or influenced by an attacker validates a matching receipt. A scope of `[ServerStatefulSet]` makes the original one-check receipt valid again.

A nonempty scope is not a correct scope. Its trust boundary does not exist yet.

## What is needed

- [ ] An immutable, controller-authenticated **creation manifest**, recorded before or while backend resources are created, listing what was actually created for this exact instance.
- [ ] Manifest entries must carry **concrete resource identity**, not categories. `TeardownSubject::Database = Verified` does not bind to a specific database; the same holds for volumes, secrets, and CIDR claims. Needed: captured PV/volume IDs, the exact datastore and database identity, secret names and UIDs, namespace, and multiplicity.
- [ ] Optional footprints that were never created are **absent** from the manifest rather than excused inside the receipt — that is what makes `NotApplicable` unnecessary for required subjects.
- [ ] Bind each receipt to the same teardown **attempt**: `completed_at` currently accepts any string and establishes neither freshness nor ordering. An attempt nonce/generation, and a check that evidence was produced after that attempt began.

## Eligibility gaps from the same review

`verified_destroy_eligibility` catches useful exclusions but is neither sound nor complete:

- [ ] `VerifiedDestroyIneligible::UnsupportedBackend` is **unreachable** — the function takes no backend argument, so a non-k3s caller can receive `Ok(())`.
- [ ] `external_datastore_identity_recorded: bool` is an assertion, not provenance. It cannot distinguish embedded from external storage, identify which database was used, or fail closed on uncertainty.
- [ ] Dynamic-storage eligibility depends on live StorageClass/PV facts (reclaim policy, bound PV identity) that are neither an input nor captured into the manifest. A lookup error, a changed StorageClass, or an unrecorded PV must make the pool ineligible.
[...]

Issue #70, opened 2026-08-10
Epic: Agent Sandbox — lease agent environments on management or child clusters


jleni:
## Goal

Make upstream [Agent Sandbox](https://github.com/kubernetes-sigs/agent-sandbox) a first-class Kobe execution kind that can be leased either:

- directly in the Kobe management cluster; or
- inside an exclusive Kobe-created child cluster.

The user-facing model is one pool/lease CLI whose resource kind determines the capabilities and implementation:

```text
Pool -> Lease -> kind-specific capabilities
```

For this epic, `SandboxPool` is the operator-defined class: placement, upstream template/warm pool, resources, isolation tier, TTL limits, and capabilities. `SandboxLease` is disposable compute. `Execution` is the agent-native command/log/cancel interface; interactive attach is optional human-facing sugar.

## Product contract

- Humans and agents use the same lease substrate.
- Agents can create, execute, observe, and release without Kubernetes credentials or an interactive terminal.
- Callers select an administrator-owned pool; they cannot submit arbitrary Pod specs, host mounts, RuntimeClasses, or target-cluster credentials.
- The first release is ephemeral-only; persistent data belongs to a later Workspace abstraction.
- Direct and child placement expose the same public API and CLI behavior.
- A child cluster provides lifecycle and blast-radius containment; it is not automatically a kernel isolation boundary.
- A configured isolation tier must be verified. Kobe must fail closed rather than silently fall back to `runc`.
- Release and expiry revoke access first, then verify cleanup. Uncertain capacity is quarantined, not re-leased.

## Child issues

Foundations:

- [x] #71 — typed `SandboxPool` / `SandboxLease` APIs and kind-aware policy
- [x] #72 — pinned external-install validation shipped; managed/child bootstrap remains part of #74's explicit approval gate

Placement and lifecycle:

- [x] #79 — UID-fenced `ClusterLease` / `ClusterInstance` binding and backend resolution
- [x] #80 — verified k3s teardown receipts and quarantine
- [x] #73 — management-cluster `SandboxClaim` reconciliation
- [ ] #74 — composition through an owned, receipt-required child `ClusterLease`, including approved runtime bootstrap

Access and closure:

- [x] #75 — agent-native Execution/access parent; decomposed by #81–#84
- [ ] #76 — live dual-placement conformance and cleanup/quarantine gate

Hardening and proof:

- [x] #100 — admission ledger namespace policy, RBAC, shape validation, and aggregate quota
[...]

PR #115, opened 2026-08-12
lifecycle(k3s): verified teardown — contract, k3s provider, and the finalizer gate

### LINKED ISSUE #80: lifecycle(k3s): verify teardown receipts and quarantine uncertain capacity
## Goal

For receipt-required k3s leases, make teardown a verified exit gate rather than a successful DELETE request.

Release/expiry revokes access first. Kobe removes the final cleanup handle only after a typed, durable receipt proves that the exact lease-owned k3s footprint is absent. Missing or uncertain evidence quarantines the unit and prevents reuse.

This is the first bounded provider for #70's child-cluster cleanup contract. It intentionally does not claim verified teardown for every Kobe backend.

Depends on the UID-fenced binding issue. Blocks #74 and the child-placement leg of #76.

## Current gap

As of `origin/main@3ce270d`:

- `ClusterBackend::delete` returns only `Result<()>`, not cleanup evidence.
- k3s PVC-list/delete, PDB-delete, Pod force-delete, and PostgreSQL-drop failures can warn and continue.
- k3s does not wait for every deleted object and backing volume to disappear before returning `Ok(())`.
- the instance controller deletes the `ClusterInstance` after `delete()` returns;
- the lease controller treats `ClusterInstance` 404 as recycling complete;
- neither `ClusterLease` nor `ClusterInstance` has a durable teardown receipt or Quarantined phase.

Therefore "instance CR is gone" does not prove workload, data, credentials, or backing storage are gone.

## Compatibility and activation

- Add a typed receipt-required cleanup mode, equivalent to `VerifiedDestroy`.
- #74-created internal `ClusterLease` objects must request this mode.
- Existing ordinary ClusterLease objects retain their current default behavior in this issue; making verified teardown universal belongs in broader fleet lifecycle work.
- Only `backend.type=k3s` implements `VerifiedDestroy` initially.
- Unsupported backends or k3s configurations are rejected before binding; they must not silently degrade to legacy cleanup.

## Initially supported k3s profile

A receipt-required k3s pool is eligible only when its footprint is verifiable:

- destroy-and-replace lifecycle, not in-place reset;
- `kubeletSharedMount` disabled unless a future host-reaper acknowledgement is part of the receipt;
- diagnostics/S3 bundle capture disabled unless its object deletion becomes receipt-verifiable;
- storage is either non-persistent or dynamically provisioned with a supported `Delete` reclaim path whose PVC and backing PV absence can be observed;
- datastore is embedded, or the exact external PostgreSQL database identity and authorized connection needed to verify its absence are recorded at creation.

Known `Retain` storage, an unobservable backing volume, missing datastore provenance, or an unsupported footprint makes the pool ineligible for receipt-required leases.

## Teardown contract

For the exact UID-fenced binding:

1. Deny new access, close active streams where applicable, and explicitly delete the lease connect token.
2. Persist the teardown attempt and exact subject UIDs before destructive work.
3. Delete the exact k3s footprint idempotently.
4. Poll and verify absence; an accepted DELETE is not evidence of completion.
5. Persist a verified receipt before releasing the instance finalizer or deleting the last retry handle.
6. Allow the owning higher-level controller to acknowledge/read the receipt before it can disappear.

The verified inventory includes, where applicable:

- agent Deployment;
- server StatefulSet and all labeled server/agent Pods;
- Service and PodDisruptionBudget;
- publisher and registries ConfigMaps;
- token and kubeconfig Secrets;
- lease connect-token Secret;
- per-instance `CIDRClaim`;
- every `data-{instance}-server-{ordinal}` PVC and its captured backing PV;
- the exact `k3s_<instance>` PostgreSQL database.

Verification uses captured UIDs and exact labels/names. A same-named replacement is a mismatch, not successful absence. `NotApplicable` is allowed only from recorded creation-time configuration; an API/RBAC/provider error is `Unknown`.

## Receipt contract

Persist a typed, non-secret record containing at least:

- schema version and attempt ID;
- lease, instance, and pool names and UIDs;
- backend type and immutable provenance/spec digest;
- started/completed timestamps;
- one result per required check: `Verified`, `NotApplicable`, or `Unknown`;
- bounded reason codes and retry count;
- aggregate outcome.

The receipt must remain queryable after the `ClusterInstance` disappears and until #74's owning `SandboxLease` has consumed or acknowledged it. An Event or log line is not a receipt.

Do not record credentials, Secret contents, command data, tenant output, database URLs, or raw provider responses.

## Quarantine contract

- Any `Unknown`, UID mismatch, unsupported footprint discovered after handout, verification timeout, or cleanup failure transitions the exact lease/instance to Quarantined.
- Quarantine is visible in typed status, conditions, pool counts, metrics, and CLI/API terminal-state handling.
- Access remains revoked.
- Finalizers, binding provenance, and retry handles remain.
- Quarantined capacity is never Ready, never rebound, and never treated as clean replacement capacity.
- Reconciliation retries idempotently with bounded backoff.
- It may leave Quarantined only after the same exact subject produces a fully verified receipt.
- No default backend, name-only cleanup, force-finalizer removal, or "warn and continue" path may manufacture a successful receipt.

## Acceptance criteria

- [ ] `VerifiedDestroy` is rejected before binding for every non-k3s backend and unsupported k3s footprint.
- [ ] #74 can request `VerifiedDestroy` without exposing this internal ClusterLease to its caller.
- [ ] Successful k3s teardown produces a receipt bound to the exact lease/instance/pool UIDs.
- [ ] The receipt proves absence of every applicable management resource, PVC/PV, credential, CIDR claim, and PostgreSQL database.
- [ ] StatefulSet/Pod/PVC deletion is polled to confirmed absence; resources recreated during teardown are detected.
- [ ] PVC/PDB/Pod/database errors that currently warn and continue instead quarantine and remain retryable.
- [ ] `reclaimPolicy=Retain`, a still-present PV, unknown volume identity, or provider API failure cannot produce a verified receipt.
- [ ] A configured PostgreSQL database is confirmed absent; missing/rotated/unreachable datastore access is uncertainty, not `NotApplicable`.
- [ ] Release and expiry use the same revoke -> destroy -> verify -> receipt path.
- [ ] Restart/failure injection after every teardown step resumes without losing the binding, inventory, attempt, or evidence.
- [ ] Instance/lease/pool name reuse and UID tampering cannot satisfy another unit's receipt.
- [ ] Quarantined capacity cannot be selected, marked Ready, or silently deleted.
- [ ] Existing non-receipt ClusterLease behavior remains backward-compatible.
- [ ] A deterministic k3s integration test covers both the verified path and representative quarantine failures.
- [ ] Generated CRDs, public status models, CLI terminal-phase handling, metrics, and operator documentation are updated and checked for drift.

## Non-goals

- Verified providers for k0s, vcluster, vkobe, or CAPI
- Direct-management Agent Sandbox workload cleanup; #73 owns that path
- In-place wipe/reset or proving reusable shared capacity
- A generic orphan-resource reaper
- Cryptographic media erasure or automatic cleanup of `Retain` volumes
- Host-mount acknowledgement or S3 diagnostics lifecycle; those configurations are initially unsupported
- A manual waiver that reports uncertain cleanup as verified
- The generic backend contract/capability matrix in #11

## Dependencies

- Depends on: the UID-fenced ClusterLease<->ClusterInstance binding issue
- Blocks: #74 and the child-cluster cleanup leg of #76
- Parent: #70
- Related broader lifecycle work: #24 and #11

emmanuelm41:
Closes #80 · follows #96 (merged as `a056f32`)
Epic: #70 — Agent Sandbox (lane `S1`)

**Targets `epic/agent-sandbox`.** v0.39.0 shipped with zero sandbox/teardown files — what holding this epic off `main` was for.

> This PR grew to cover all three remaining layers of #80. Earlier descriptions covered only the contract.

## The problem

`ClusterBackend::delete` returns `Result<()>`. The instance controller releases the cleanup finalizer the moment it returns Ok, and the lease controller reads the resulting 404 as "recycling complete". So **an accepted DELETE request was proof of destruction** — for capacity that held another tenant's code and credentials. PVC, PDB, Pod and PostgreSQL failures warn and continue, and nothing waits for objects or their backing volumes to disappear.

## Three layers

**1 · Contract.** `delete_verified` is a **new trait method with a default body**, not a changed `delete` signature. `ClusterBackend` has five implementors; changing the signature forces a mechanical update to all five, and the mechanical update is `Ok(vec![])` — a backend claiming a clean footprint it never inspected. Refusing by default means a backend can only claim evidence by writing the code.

**2 · k3s provider.** Capture → delete → **poll to confirmed absence**. Capture runs first because a PVC's `spec.volumeName` is the only link to its backing PV and it vanishes with the PVC; capture late and you can prove the claim is gone but never the volume. **Only a 404 counts as proof** — a 403 or a timeout becomes `Unknown`, so a broken RBAC grant can never read as a clean teardown.

**3 · Finalizer gate.** Release now requires a receipt covering the plan recorded at creation. Anything unproven **quarantines and keeps every handle** — finalizer, binding, provenance — because those are what a retry needs to address the same exact subject, and what stops the 404-based recycle path from treating the capacity as clean.

## The trust boundary #102 asked for

The plan is **stamped into immutable provenance at creation**, not derived at teardown. Deriving it later reads whatever the config says *then* — so a pool that stopped setting `registryMirrors` after an instance was built would silently drop that ConfigMap from the plan and never verify it. A receipt can no longer define its own scope.

Paired footprints are planned together: persistent storage pulls in the PVC **and** its volume; a datastore pulls in the database **and** its role. Proving one half while the other survives is exactly the leak a receipt would hide.

An instance created before verified teardown existed has no plan and is quarantined rather than verified against a guess.

The receipt is persisted **before** anything is released — written afterwards it could be lost with the object it describes, and outliving that object is its whole purpose.

## Caught by keeping the epic branch synced

**`TeardownSubject::DatabaseRole`.** `fix(datastore): give each cluster its own PostgreSQL role` (#105) landed on `main` while this branch was separate. k3s teardown now drops a **role** as well as the database — all best-effort `warn`-and-continue. Layer 1's subject list predated it, so a receipt could have proven the database absent **while a credential-bearing role survived**.

**`UnsupportedBackend` was unreachable** — eligibility took no backend to judge, so k0s/vcluster pools got `Ok(())`. Now refused.

## Regression proof

Mutation-checked at the site each fix lives:

| Mutation | Caught by |
|---|---|
| treat a failed lookup as absence | `a_failed_lookup_is_uncertainty_not_absence` |
| ignore a failed delete | `a_failed_delete_makes_every_subject_unknown` |
| plan a PVC without its volume / a DB without its role | `paired_footprints_are_never_planned_alone` |

Plus: surviving object → `Unknown`; missing object → `Verified`; database and role `Unknown` without a datastore; uncapturable volumes fail closed; plan covers exactly what a config creates.

## Validation

```
cargo test --workspace                                  # 1351 passed, 0 failed
crdgen drift                                            # clean
```

## Scope, stated narrowly

Leases without `cleanupMode: VerifiedDestroy` take the **unchanged** path — the gate returns before touching anything.

Still open and tracked, not silently skipped:
- **#103** — quarantine is not surfaced in pool phase, status, CLI or metrics, and has no receipt-gated exit.
- **#102** — subjects are categories, not concrete resource identities: one `Database = Verified` does not bind to a specific database.

### DIFF
--- a/charts/kobe/crds/clusterinstances.yaml
+++ b/charts/kobe/crds/clusterinstances.yaml
@@ -927,6 +927,46 @@ spec:
                       not infer it from a same-named current pool.
                     nullable: true
                     type: string
+                  teardownPlan:
+                    description: |-
+                      Exactly which footprints this instance created, recorded at creation.
+
+                      This is the trust boundary for verified teardown. A receipt only proves
+                      what it covers, so the list of subjects it must cover cannot come from
+                      the receipt, nor from whatever the teardown path happens to look for
+                      later — an optional resource whose config changed after creation would
+                      then be quietly dropped from the plan and never verified.
+
+                      Stamped once alongside the rest of this provenance and never
+                      overwritten. `None` means the instance predates verified teardown; such
+                      an instance is ineligible for [`CleanupMode::VerifiedDestroy`] rather
+                      than being verified against a guessed plan.
+                    items:
+                      description: |-
+                        One thing whose absence must be proven.
+
+                        Enumerated rather than free-text so a receipt cannot be satisfied by an
+                        unrecognised or invented subject, and so adding a resource to the teardown
+                        path is a deliberate, reviewable change to this list.
+                      enum:
+                      - agentDeployment
+                      - serverStatefulSet
+                      - serverPods
+                      - service
+                      - podDisruptionBudget
+                      - publisherConfigMap
+                      - registriesConfigMap
+                      - tokenSecret
+                      - kubeconfigSecret
+                      - cidrClaim
+                      - connectTokenSecret
+                      - serverDataPvcs
+                      - serverDataVolumes
+                      - database
+                      - databaseRole
+                      type: string
+                    nullable: true
+                    type: array
                 required:
                 - operatorVersion
                 type: object
--- a/charts/kobe/crds/clusterinstances.yaml
+++ b/charts/kobe/crds/clusterinstances.yaml
@@ -1107,6 +1107,22 @@ spec:
                   timestamp fields above.
                 nullable: true
                 type: string
+              teardownIdentities:
+                description: |-
+                  Provisioner-assigned resource identities this instance owns — currently
+                  the PersistentVolumes its data claims are bound to.
+
+                  Captured once the instance is healthy, NOT at teardown. These names are
+                  chosen by the provisioner and only knowable after claims bind, so
+                  capturing them during teardown would let the teardown path define its
+                  own scope. Recorded early, they become something a later teardown must
+                  account for rather than something it gets to choose.
+
+                  Written once and never shrunk: a smaller list would silently narrow what
+                  a receipt has to prove.
+                items:
+                  type: string
+                type: array
             type: object
         required:
         - spec
--- a/charts/kobe/crds/clusterleases.yaml
+++ b/charts/kobe/crds/clusterleases.yaml
@@ -368,6 +368,7 @@ spec:
                           - serverDataPvcs
                           - serverDataVolumes
                           - database
+                          - databaseRole
                           type: string
                       required:
                       - result
--- a/charts/kobe/crds/clusterleases.yaml
+++ b/charts/kobe/crds/clusterleases.yaml
@@ -370,6 +370,23 @@ spec:
                           - database
                           - databaseRole
                           type: string
+                        verified:
+                          description: |-
+                            The concrete resources this check actually looked at.
+
+                            A [`TeardownSubject`] is a category. On its own,
+                            `Database = Verified` asserts that *a* database is gone without saying
+                            which — so a receipt could be satisfied by checking the wrong one, or by
+                            checking one of several when the footprint had more. Recording the exact
+                            identities makes the claim auditable after the fact, which is the whole
+                            point of keeping evidence that outlives the instance.
+
+                            Names only: object names, PV names, the database and role identifiers.
+                            Never connection strings, credentials, or anything that would turn a
+                            receipt into a secret.
+                          items:
+                            type: string
+                          type: array
                       required:
                       - result
                       - subject
--- a/charts/kobe/crds/clusterpools.yaml
+++ b/charts/kobe/crds/clusterpools.yaml
@@ -916,6 +916,18 @@ spec:
                 - Idle
                 nullable: true
                 type: string
+              quarantined:
+                default: 0
+                description: |-
+                  Members held back because their teardown could not be proven complete.
+
+                  Separate from `recycling` on purpose: recycling is transient churn that
+                  resolves itself, while a quarantined member is stuck until the same
+                  exact subject produces a verified receipt. Reporting them together would
+                  make a pool bleeding capacity look merely busy.
+                format: uint32
+                minimum: 0.0
+                type: integer
               queueDepth:
                 default: 0
                 description: Current queue depth (claims waiting for clusters).
--- a/src/backend/datastore.rs
+++ b/src/backend/datastore.rs
@@ -461,6 +461,41 @@ pub async fn ensure_cluster_role(
 
 /// Drop the per-cluster role. Call AFTER `drop_database`: PostgreSQL refuses to
 /// drop a role that still owns objects.
+/// Whether this cluster's database still exists.
+///
+/// Verification, not cleanup: `DROP DATABASE IF EXISTS` succeeding proves the
+/// statement ran, not that the database is gone — a concurrent recreate, a
+/// failed drop that only warned, or a connection to a different server all
+/// leave it present. Teardown evidence has to come from a separate read.
+///
+/// An error is deliberately propagated rather than reported as "absent". A
+/// query that could not run is uncertainty, and uncertainty must quarantine.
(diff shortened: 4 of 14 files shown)


emmanuelm41:
CI is red on **runner infrastructure, not this change**.

Every failure is in `Set up Bun for E2E harness tests`, before any Rust compilation:

```
           http2 error: stream error received: refused stream
curl: (56) Connection died, tried 5 times before giving up   # fetching the mise tarball itself
```

Evidence it is environmental:

- **#115 and #116 fail identically** and share no code — #115 is teardown, #116 is `src/api/sandbox.rs`.
- An earlier run on #115's own branch **succeeded at 18:58**; the failures start after ~19:00.
- It never reaches `cargo`; the step that fails only downloads bun and mise.

Verified locally on both branches: `cargo fmt --check`, `clippy --workspace --all-targets -D warnings`, and the full `cargo test --workspace` suite all clean, plus CRD drift.

Rerun once per the transient-failure rule; it failed the same way, so I have stopped rather than blind-retrying a network fault. Needs a re-run once runner egress recovers — or a cached/pinned bun if this keeps recurring, since a CDN hiccup currently fails the whole Rust gate.

Issue #101, opened 2026-08-11
hardening(sandbox): durable reaper and an honest failure contract for admission


emmanuelm41:
Found by cross-family review of PR #93 (the #71 admission ledger). PR #93 fixed the user-visible lockout; these are the residual gaps it deliberately did not close.

## 1. The reaper is opportunistic, not durable

PR #93 reclaims **this principal's** abandoned unadmitted leases **at create time only**, so a caller locked out by their own crashed request recovers on their next attempt. A principal who never calls create again keeps their slot and alias stranded indefinitely.

- [ ] A controller-owned sweep that reclaims abandoned pending leases regardless of caller activity.

## 2. Wall-clock age is not proof of death

The sweep treats a `pending` lease older than 10 minutes as abandoned. The fencing is sound — a reap and a concurrent admission cannot both win — but age is an availability heuristic, not a protocol guarantee. A paused, starved, or network-stalled request can still legitimately own a lease past the deadline and then fail, having done everything right.

- [ ] Bound and cancel admission before the reap deadline, or give the handler a server-owned admission lease / heartbeat whose expiry actually proves no request can still commit.

## 3. Failure signal does not match cluster state

Several paths return 503 to the caller while leaving an **admitted** lease behind:

- the PATCH commits but the response is lost → GET still reads `pending` → the fenced DELETE 409s against the late-landing patch → returns `Kubernetes(_)`, not `AdmissionNotCommitted`, so the handler returns 503 with no cleanup;
- `validate_lease_shape` failing on the admit response (`LeaseShapeChanged` — any mutating webhook or CRD spec defaulting makes `actual.spec != expected.spec` structurally) would 503 on *every* create while still admitting the lease.

Quota accounting stays consistent, but the caller is told it failed and will retry — ending with two live sandboxes for one intended request.

- [ ] Classify these outcomes so the response matches what actually exists.

## 4. Smaller items from the same review

- [ ] `uid: created.uid().unwrap_or_default()` — an empty UID makes every later release 409, silently swallowed as "already gone", stranding the reservation. Should hard-fail like the neighbouring `MissingLeaseUid`.
- [ ] A lease whose admission annotation is missing or corrupted is undeletable through the API: the release path routes any non-`admitted` value into a delete that demands exactly `pending`, so anything else 503s forever.
- [ ] `requester_type` is stored but omitted from both `principal_hash` and `principal_matches`, so identities differing only by that field share one quota and alias namespace. Confirm that is intended.
[...]

Issue #100, opened 2026-08-11
security(sandbox): fence the admission ledger against direct Kubernetes access


emmanuelm41:
Found by cross-family review of PR #93 (the #71 admission ledger). Not a defect in that PR's logic — a set of assumptions it silently depends on, which nothing currently enforces.

## The dependency

Admission enforces per-principal quota and alias uniqueness by `CREATE`-ing coordination `Lease` objects at deterministic names as compare-and-swap tokens. That is only an abuse boundary while **the reservation namespace is writable exclusively by the operator's service account**. Today nothing states or checks that.

If any tenant workload or caller can create/patch/delete coordination Leases in that namespace, admission is trivially bypassable:

- create the exact quota names for a victim's principal → block them entirely;
- more cheaply, create *any* objects carrying the two labels the advisory LIST matches → the handler refuses before its authoritative scan, no valid names or owner references needed;
- delete a live reservation while its `SandboxLease` is still admitted → the slot frees while the lease persists, so the next request over-admits;
- create the exact alias names → squat aliases.

Predictable names plus label selectors make this cheap. A label selector is not an integrity check.

## Confused-deputy delete

`release_reservations_for_lease` lists every coordination Lease whose lease-UID label matches, then deletes them **without validating** reservation type, expected name, principal, annotation, or owner reference. An actor who can create a Lease in that namespace can label it with a victim `SandboxLease` UID; Kobe's privileged credentials then delete that injected object on the victim's cleanup.

## Scope

- [ ] Document and enforce that the reservation namespace is operator-SA-writable only; no tenant identity gets Lease write/delete there.
- [ ] Validate the full expected reservation shape before deleting — type label, principal label, owner reference, and a name matching the deterministic scheme — rather than trusting a bare label match.
- [ ] Prefer computing expected reservation names from persisted admission metadata over deleting arbitrary label matches.
- [ ] Add a namespace-level object quota so reservation growth is bounded across identities, not only per principal.

## Non-goals

Changing the CAS mechanism itself; it is correct given the RBAC assumption. This issue makes the assumption explicit and checked.

emmanuelm41:
The RBAC half is code after all — the Helm chart is in this repo. I was wrong to call it cluster configuration. Fixed in PR #116, and it turned up **two shipping-blockers** against work earlier in this epic:

[...]

Issue #73, opened 2026-08-10
controller(sandbox): reconcile SandboxClaim leases in the management cluster


jleni:
## Goal

Implement direct `SandboxLease` placement in Kobe's management cluster using an administrator-controlled upstream `SandboxTemplate` / `SandboxWarmPool` and one `SandboxClaim` per lease.

Depends on #71 and #72. This is a child of #70; isolation qualification remains related to #14.

Direct placement is opt-in. Callers select a Kobe `SandboxPool`; they cannot choose upstream objects, Pod specs, host mounts, runtime settings, namespaces, or management credentials.

## Reconciliation contract

- Reconcile each direct `SandboxPool` into controller-owned upstream Template/WarmPool objects in an operator-owned pool namespace.
- Reconcile each `SandboxLease` into one Claim in that same pool namespace.
- Record and revalidate exact namespace, Claim, Sandbox, Pod, Template, and WarmPool names **and UIDs**; names alone are not sufficient authority.
- Gate pool readiness on the expected runtime, node placement, limits, Pod Security, network isolation, controllers/webhook, and a real execution canary.
- Mark plain runc capacity honestly as `trusted-runc`; never represent it as hostile-code isolation.
- Start lease TTL only after the Claim/Sandbox is Ready; bound setup with a separate provisioning deadline.
- Set the upstream absolute shutdown time and `DeleteForeground` as a backstop to Kobe reconciliation.
- Keep management kubeconfigs, service-account tokens, and the unrestricted Kubernetes/Agent Sandbox APIs internal.

## Teardown contract

Release/expiry must:

1. reject new access and close active streams;
2. foreground-delete the Claim;
3. verify the lease-owned Sandbox, Pod, Service, PVC/PV/backing volume, and access RBAC are absent;
4. verify the warm pool has replenished only clean Ready capacity;
5. remove the Kobe finalizer and report Released/Expired.

The pool namespace and controller-owned Template/WarmPool are pool resources and remain until the pool is deleted. UID mismatch, unreachable API, broken webhook, stuck deletion, or uncertain storage cleanup quarantines the referenced capacity and fails closed.

## Acceptance criteria

- [ ] Pool and lease reconciliation is idempotent across restarts and duplicate events.
- [ ] A Ready direct pool can create a Claim and expose a Ready `SandboxLease` with exact immutable provenance.
- [ ] Unsupported/missing RuntimeClass, unsafe policy, failed canary, or ambiguous ownership makes the pool unready before handout.
[...]

Issue #72, opened 2026-08-10
dist(agent-sandbox): install a pinned runtime in management and child clusters


jleni:
## Goal

Make upstream Agent Sandbox reproducibly available wherever Kobe may place a `SandboxLease`:

1. the Kobe management cluster; and
2. Kobe-created child clusters.

This is a child of #70.

## Scope

- Add one Helm mode with explicit values equivalent to `disabled | managed | external`.
- In `managed` mode, install the required Agent Sandbox core and extensions CRDs, controllers, and webhook because Kobe consumes `SandboxTemplate`, `SandboxWarmPool`, and `SandboxClaim`.
- In `external` mode, own none of those resources and validate that the installed APIs/version are compatible.
- Pin one released upstream version and immutable image/artifact references; never fetch mutable `main`/`latest` manifests at runtime.
- Refuse ambiguous or conflicting installation ownership on the same Kubernetes API.
- Add a built-in `BootstrapConfig` path that installs the same pinned release in a newly leased child cluster.
- Make both paths idempotent and gate readiness on API discovery, controller/webhook health, and a real create/delete claim canary.
- Define explicit CRD upgrade/retention behavior. Uninstall must not silently delete user data or externally owned CRDs.
- Keep the optional Agent Sandbox Router disabled.

A chart dependency is preferred if upstream publishes a suitable release chart; otherwise vendor release-generated manifests with provenance.

## Acceptance criteria

- [ ] Disabled mode renders no Agent Sandbox resources and leaves current Kobe behavior unchanged.
- [ ] Managed mode installs the pinned core + extensions release on a fresh supported cluster.
- [ ] External mode succeeds for a compatible installation and fails clearly for missing/incompatible APIs.
- [ ] Management and child installation use the same pinned version and immutable artifacts.
- [ ] Child bootstrap is safe to retry after Kobe/controller restarts.
- [ ] Installation is not Ready until a minimal `SandboxTemplate` / `SandboxWarmPool` / `SandboxClaim` create-and-delete canary succeeds.
- [ ] Conflicting install ownership, incompatible CRDs, broken webhook, or failed cleanup reports a fail-closed condition.
- [ ] Helm tests cover disabled, managed, external, and invalid configurations.
- [ ] No Router Deployment, Service, endpoint, or Router RBAC is installed.
- [ ] Public operator documentation covers modes, prerequisites, version compatibility, upgrades, and uninstall behavior.

## Non-goals

[...]

PR #116, opened 2026-08-12
security(sandbox): validate reservation shape, and reap abandoned admission state

emmanuelm41:
Closes the delete-side of #100 · closes the durable-reaper half of #101
Epic: #70 — Agent Sandbox

**Sibling of #115** — that one is teardown, this is `src/api/sandbox.rs`; no dependency between them.

Both fixes come from cross-family review of #93 (Codex, plus a same-lineage Claude agent).

## 1 · Confused-deputy delete (#100)

`release_reservations_for_lease` listed every coordination `Lease` carrying a matching lease-UID label and deleted **all of them**.

A label is caller-supplied data, not an integrity check. Anyone able to create a coordination Lease in that namespace could stamp a victim's `SandboxLease` UID onto an unrelated object — say `kube-controller-manager` — and Kobe's privileged credentials would delete it during the victim's ordinary cleanup. The attacker needs no permission Kobe has; they borrow Kobe's.

The selector now only **narrows the search**. What may be deleted is decided from names derived from *this lease's own principal*: names are deterministic (`sbx-quota-{hash}-{slot}`, `sbx-alias-{hash}-{alias}`), so anything not matching either shape is not ours, however it is labelled. A crafted suffix like `...-0-extra` does not match either.

Cleanup runs from the persisted lease rather than a live request, so this adds `principal_hash_for` over the stored `SandboxPrincipal` — with a test pinning that it agrees with the live-identity digest. Had they diverged, cleanup would compute names that never match what admission created and silently release nothing, stranding quota on *every* teardown.

## 2 · Durable reaper (#101)

The create-path sweep only helps a caller who comes back. A principal whose request died between reserving and admitting, and who never calls create again, kept their slot and alias consumed **indefinitely** — invisibly, because the lease looks like an ordinary `pending` object no controller will touch.

`run_sandbox_admission_reaper` runs on the operator, so recovery no longer depends on the victim retrying. Same fences throughout: only leases past the deadline, deleted under UID + resourceVersion preconditions, releasing only reservations whose names derive from that lease's principal.

It deliberately does **not** depend on Sandbox placement (#73) — it touches only admission objects, so it ships while placement is blocked behind #72.

## 3 · Empty-UID reservations

`created.uid().unwrap_or_default()` in **both** acquire arms. An empty UID makes every later release fail its precondition with 409, which the release path treats as "ours is already gone" — silently stranding the very slot the reaper exists to protect. Now rolls back and fails: acquiring something we can never release is worse than refusing.

The review named only the alias arm. The quota arm had the identical defect one function below — fixing only what was reported would have left the more common path broken.

## Regression proof

- `only_names_derived_from_our_principal_may_be_deleted` — ours accepted; injected `important-system-lease` refused; another principal's refused; right-prefix-wrong-shape refused. **Mutation-checked**: restoring "accept any name" fails it.
- `stored_and_live_principal_digests_agree` — catches the silent-no-op divergence.
- `the_reaper_reclaims_a_lease_whose_principal_never_returns` — the case the create-path sweep cannot cover.

## Validation

```
cargo test --workspace                                  # 1334 passed, 0 failed
```

> CI is currently red on **runner infrastructure** — `mise install bun` fails to download before any Rust compilation, identically on #115 which shares no code. See the comment below.

## Scope, stated narrowly

Still open in **#100**: RBAC so no tenant identity can create or delete coordination Leases in the reservation namespace at all, plus a namespace object quota. Operational configuration, not code — but load-bearing, since the CAS ledger is only an abuse boundary while that RBAC holds.

Still open in **#101**: paths returning 503 while leaving an admitted lease behind; a corrupted admission annotation making a lease undeletable through the API; `requester_type` absent from the principal digest; per-principal rate limiting.

### DIFF
--- a/charts/kobe/templates/rbac.yaml
+++ b/charts/kobe/templates/rbac.yaml
@@ -92,10 +92,26 @@ rules:
   - apiGroups: [""]
     resources: ["nodes"]
     verbs: ["get", "list", "watch"]
-  # Leader election
+  # Verified teardown must observe that a PVC's backing PersistentVolume is
+  # actually gone — a claim can be deleted under a `Retain` reclaim policy while
+  # the volume, and the tenant's data, survive. Read-only: kobe never deletes a
+  # PV directly, it only proves absence. Without this the volume check 403s,
+  # which correctly reads as "cannot verify" and would quarantine every
+  # persistent pool forever.
+  - apiGroups: [""]
+    resources: ["persistentvolumes"]
+    verbs: ["get", "list", "watch"]
+  # Leader election, AND the Sandbox admission ledger.
+  #
+  # `list` and `delete` are load-bearing for admission, not leader election:
+  # quota and alias reservations are coordination Leases, released by deleting
+  # them and found by listing on the owning lease's UID label. Without both,
+  # every release 403s and a caller's quota is consumed permanently on the
+  # first teardown — a failure no unit test sees, because it only exists once
+  # real RBAC applies.
   - apiGroups: ["coordination.k8s.io"]
     resources: ["leases"]
-    verbs: ["get", "create", "update", "patch"]
+    verbs: ["get", "list", "create", "update", "patch", "delete"]
   # CRD discovery (Velero detection)
   - apiGroups: ["apiextensions.k8s.io"]
     resources: ["customresourcedefinitions"]
--- a/src/api/sandbox.rs
+++ b/src/api/sandbox.rs
@@ -1030,7 +1030,12 @@ async fn admit_sandbox_lease(
                     validate_lease_shape(lease, &current, SANDBOX_ADMISSION_ADMITTED)
                 }
                 Some(SANDBOX_ADMISSION_PENDING) => {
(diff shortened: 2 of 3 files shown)


emmanuelm41:
CI is red on **runner infrastructure, not this change**.

Every failure is in `Set up Bun for E2E harness tests`, before any Rust compilation:

```
           http2 error: stream error received: refused stream
curl: (56) Connection died, tried 5 times before giving up   # fetching the mise tarball itself
```

Evidence it is environmental:

- **#115 and #116 fail identically** and share no code — #115 is teardown, #116 is `src/api/sandbox.rs`.
- An earlier run on #115's own branch **succeeded at 18:58**; the failures start after ~19:00.
- It never reaches `cargo`; the step that fails only downloads bun and mise.

Verified locally on both branches: `cargo fmt --check`, `clippy --workspace --all-targets -D warnings`, and the full `cargo test --workspace` suite all clean, plus CRD drift.

Rerun once per the transient-failure rule; it failed the same way, so I have stopped rather than blind-retrying a network fault. Needs a re-run once runner egress recovers — or a cached/pinned bun if this keeps recurring, since a CDN hiccup currently fails the whole Rust gate.

Issue #84, opened 2026-08-10
cli(sandbox): add exec, logs, cancel, attach, port-forward, and run


jleni:
## Goal

Expose the Sandbox API as composable human and agent CLI commands with stable machine-readable output, exact remote exit codes, aliases, and reliable release cleanup.

This is a child of #75.

## Commands

```text
kobe sandbox exec <lease-or-alias> --output json -- <argv...>
kobe sandbox logs <lease-or-alias> --execution <id> [--follow]
kobe sandbox cancel <lease-or-alias> --execution <id>
kobe sandbox port-forward <lease-or-alias> <local>:<declared-remote>
kobe sandbox attach <lease-or-alias>
kobe sandbox run <pool> --ttl 2h --output json -- <argv...>
```

- `exec` defaults to non-TTY and sends argv without a shell.
- JSON/NDJSON have versioned stable fields and keep stdout/stderr/exit status distinct.
- `run` composes create -> wait -> non-TTY execution -> release and returns the remote exit code.
- `run` attempts release on success, remote failure, signal, timeout, network error, and client cancellation. Cleanup failure is reported separately and never hidden by the command result.
- Alias resolution is scoped to the caller and rejects ambiguity.
- `attach` is an optional TTY wrapper, not required for agents.
- No command downloads kubeconfig or exposes target credentials.

## Acceptance criteria

- [ ] Commands work identically for management and child placements.
- [ ] Machine-readable output is deterministic and documented.
- [ ] `exec`/`run` preserve the exact remote exit code while separately reporting transport/release failures.
- [ ] `run` attempts release on every terminal path and waits for observable Releasing/terminal state within a bound.
- [ ] Retry uses idempotency keys and cannot duplicate a command.
- [ ] Foreign, missing, ambiguous-alias, stale, expired, released, and quarantined targets fail closed.
- [ ] TTY resize/signals and declared-port forwarding work through the restricted API only.
- [ ] CLI tests cover signals, timeout, disconnect, non-zero exit, cancellation, and failed release.
- [ ] Public command/API documentation includes examples for headless agents.

## Non-goals

- Kubernetes kubeconfig access
- File synchronization or arbitrary secret injection
- Persistent Workspace attach
- iroh, SSH, Router, Herdr, or x402

## Dependencies

- Depends on: the non-TTY execution and bounded-stream children of #75
- Blocks: #76
- Parent: #75

jleni:
[...]

Issue #81, opened 2026-08-10
access(sandbox): resolve exact lease targets and mint scoped credentials


jleni:
## Goal

Resolve every Sandbox operation through the caller's exact, owned `SandboxLease` target and mint only short-lived, namespace-scoped credentials for that target.

This is a child of #75. It provides the common restricted access substrate consumed by execution, logs, attach, and port forwarding; it is not a general Kubernetes proxy.

## Contract

Resolve fail closed through:

```text
authenticated principal
  -> owned, Ready, unexpired SandboxLease UID
  -> immutable direct/child placement provenance
  -> exact Claim/Sandbox/Pod UIDs
  -> approved container and declared ports
```

- Direct and child adapters must return one typed target or a bounded denial reason. They never select a default cluster, namespace, Pod, container, or backend.
- Mint a lease-specific ServiceAccount/Role/RoleBinding and short-lived TokenRequest credential in the exact target namespace, or an equivalently least-privileged mechanism.
- Permit only the exact Kobe-owned Pod/container subresources needed by the operation.
- Canonicalize and allowlist methods, paths, stream protocols, headers, containers, and ports.
- Strip caller authorization, impersonation, and hop-by-hop headers before target access.
- Run a SandboxLease watch on every API replica. Release, expiry, quarantine, UID change, or target loss revokes new access and cancels that replica's active streams immediately.
- Audit principal, lease UID, operation, target UID, and outcome without credentials, stdin, command environment, stdout, or stderr.

## Acceptance criteria

- [ ] Management and child placement resolve through the same typed interface.
- [ ] Foreign identities receive 404 and cannot distinguish absent from unowned leases.
- [ ] Pending, Releasing, Released, Expired, Quarantined, malformed-expiry, missing-target, stale-UID, and provenance-drift paths deny before credential minting.
- [ ] Namespace, Claim, Sandbox, Pod, container, backend, and pool substitution tests fail closed.
- [ ] Scoped credentials cannot read Secrets, mutate RBAC, create Pods, access Nodes, or cross namespaces.
- [ ] Only approved containers and pool-declared ports resolve.
- [ ] Caller authorization, impersonation, and hop-by-hop headers are stripped.
- [ ] Release/expiry cancels already-upgraded streams on every serving replica before target cleanup.
- [ ] Credentials are short-lived and never appear in CRDs, responses, events, logs, or test artifacts.
[...]

Issue #76, opened 2026-08-10
conformance(sandbox): prove dual-placement isolation, expiry, and cleanup


jleni:
## Goal

Create one conformance suite proving that `SandboxPool → SandboxLease → Execution` has the same public lifecycle, authorization, expiry, and cleanup contract in both supported placements:

1. directly in the Kobe management cluster (#73); and
2. through an internal receipt-capable k3s `ClusterLease` (#74).

Depends on #71, #72, #73, #74, #75, #79, and #80. This is the closure gate for #70. Other child backends remain `Unsupported` until they implement equivalent teardown evidence.

## Required matrix

Run the same placement-parameterized scenarios for:

| Scenario | Management | Child k3s cluster |
|---|---:|---:|
| Install/readiness/canary | ✓ | ✓ |
| Create and become Ready | ✓ | ✓ |
| Execution result/log/cancel | ✓ | ✓ |
| TTY and declared port-forward | ✓ | ✓ |
| Explicit release | ✓ | ✓ |
| TTL expiry with active stream | ✓ | ✓ |
| Cancel while provisioning | ✓ | ✓ |
| Controller/API restart at each phase | ✓ | ✓ |
| Target loss and cleanup retry | ✓ | ✓ |
| Cross-tenant denial | ✓ | ✓ |
| Quarantine on uncertain teardown | ✓ | ✓ |

## Acceptance criteria

- [ ] A lease becomes Ready only after its upstream Claim/Sandbox and execution canary are Ready.
- [ ] Runtime TTL begins at Sandbox readiness, not request creation or child-cluster binding.
- [ ] Both placements return the same typed status and Execution behavior without target credentials.
- [ ] `kobe sandbox run` proves create → wait → exec → exact exit code → release for both placements.
- [ ] Explicit release and natural expiry reject new access and terminate active streams before destructive cleanup.
- [ ] Direct cleanup proves the lease-owned Claim, Sandbox, Pod, Service, access identity/RBAC, and execution records are absent; no prohibited lease-owned PVC/PV exists; the pool-owned namespace/Template/WarmPool remain healthy with clean capacity.
- [ ] Child cleanup consumes a verified receipt proving the internal `ClusterLease` footprint and original `ClusterInstance` UID are absent. Any replacement has a distinct UID and becomes Ready only through fresh provisioning.
- [ ] Cancelling a pending child placement releases partial cluster/bootstrap allocations.
[...]

Issue #74, opened 2026-08-10
placement(sandbox): compose child-cluster sandboxes through ClusterLease


jleni:
## Goal

Add child placement where Kobe acquires one exclusive internal `ClusterLease`, installs/verifies the pinned Agent Sandbox runtime, and creates the outer lease's `SandboxClaim` inside that exact child cluster.

Depends on #71, #72, #73, #79, and #80. This is a child of #70 and must preserve #73's public lifecycle contract. The first qualified child backend is a receipt-capable k3s pool using #80's `VerifiedDestroy` mode; every other backend must return `Unsupported` rather than silently weakening cleanup.

The internal cluster is an implementation detail. The caller receives Sandbox operations, never cluster-admin capacity.

## Composition contract

- A child-mode `SandboxLease` creates or adopts exactly one controller-owned internal `ClusterLease` from the `ClusterPool` selected by its `SandboxPool`.
- Persist non-secret provenance: child lease UID, bound `ClusterInstance` UID/generation, pool/backend identity, target namespace, and every upstream object UID.
- Resolve the internal lease and instance through #79's reciprocal UID-fenced binding. Reject changed, stale, name-reused, ambiguous, or fallback backend provenance before target mutation or deletion.
- The requester cannot connect to, extend, release, or obtain credentials for the internal `ClusterLease`.
- Read its backend kubeconfig only from controller-authorized storage and keep it in memory; never copy it into Sandbox status or API responses.
- Bootstrap #72 and verify the exact artifact/version, API compatibility, controllers/webhook, runtime capability, and a real create/delete canary before admitting the Claim.
- Size/extend the child lifetime to cover bounded provisioning + Sandbox runtime TTL + drain/cleanup grace. Reject a composition whose cluster policy cannot guarantee that bound.
- Use the same public Sandbox status/access contract as direct placement.

## Isolation contract

The pool declares the **actual** child backend and workload isolation tier. Exclusivity gives lifecycle containment; an exclusive k3s child is not automatically a VM/kernel security boundary. Hardened hostile-agent claims require a qualified tier under #14 or an appropriate VM/dedicated backend.

## Teardown contract

Release/expiry order is:

1. revoke Sandbox access and active streams;
2. foreground-delete the Claim and verify lease-owned workload/storage absence;
3. release the internal `ClusterLease` in `VerifiedDestroy` mode;
[...]

PR #117, opened 2026-08-12
dist(agent-sandbox): validate an operator-installed runtime (external mode)

### LINKED ISSUE #72: dist(agent-sandbox): install a pinned runtime in management and child clusters

emmanuelm41:
Closes #72 (narrowed scope)
Epic: #70 — Agent Sandbox (lane `F2`)

## This does not approve the paused effects — it avoids them

@jleni paused #72 pending approval of three effects. **None are implemented here.** The gate stands as written; the scope was narrowed to the part that needs no approval, on the epic owner's decision.

| Paused effect | Status |
|---|---|
| `managed` Helm installs + **retains** cluster-scoped CRDs/RBAC/webhook across uninstall | **Not implemented** |
| Validation creates/deletes a **real `SandboxClaim`** canary | **Not implemented** |
| Child bootstrap performs **cluster-admin** install | **Not implemented** |

## What ships

`agentSandbox.mode: disabled | external`, defaulting to **disabled** — an upgrade never starts depending on a runtime you haven't installed.

In `external` mode the operator installs the runtime and **Kobe owns nothing**. It verifies all three consumed CRDs are established and serve the pinned `extensions.agents.x-k8s.io/v1beta1`.

- **All three are required.** A runtime with claims but no warm pools would pass a laxer check and then fail to serve.
- **The version is pinned**, not accepted as-found. Kobe projects pool and lease config onto a specific schema; another version can accept the same object while meaning something else by it — which surfaces as a claim no controller reconciles rather than as an error.
- **Validated at startup**, not per request. A cluster that looks healthy and fails only when someone first uses Sandbox is the worse failure. The error names the version actually installed.

## `managed` fails loudly

Refused with an explicit error rather than degrading to `external` or `disabled`. Falling back would let a config value quietly select behaviour nobody approved — and would make the approval look granted to anyone reading the running config. Case-insensitive, and an unrecognised value is refused too rather than silently defaulting.

The refusal lives in the operator, not the chart template, so a direct (non-Helm) deployment can't route around it.

## An honest limit

**Presence and version are a weaker guarantee than a canary.** A runtime whose controller is wedged passes this and fails later. That is the direct cost of not writing a live `SandboxClaim`, and it's recorded in the module docs rather than papered over. Worth revisiting if the canary is ever approved.

## Validation

```
cargo test --workspace                                  # 1336 passed, 0 failed
helm template (default + external)                      # renders; AGENT_SANDBOX_MODE plumbed
```

## What this unblocks

#73 → #74 → #81–#84 → #76, for clusters where the runtime is operator-installed. `managed` mode can follow as its own change once the retention question is settled — the Helm `crds/` one-way door deserves its own decision, not a side effect of unblocking an epic.

> CI is currently red on runner infrastructure (`mise install bun` fails before any Rust compilation), unrelated to this change.

### DIFF
--- a/charts/kobe/templates/deployment.yaml
+++ b/charts/kobe/templates/deployment.yaml
@@ -68,6 +68,11 @@ spec:
               value: {{ .Values.logLevel | quote }}
             - name: KOBE_SYNC_IMAGE
               value: {{ include "kobe.syncImage" . | quote }}
+            # `managed` is rejected by the operator at startup rather than
+            # here, so the refusal and its reason live in one place and a
+            # direct (non-Helm) deployment cannot route around it.
+            - name: AGENT_SANDBOX_MODE
+              value: {{ .Values.agentSandbox.mode | default "disabled" | quote }}
             - name: POOL_SERVICE_ACCOUNT
               value: {{ .Values.poolServiceAccount | default (include "kobe.serviceAccountName" .) | quote }}
             - name: POOL_PUBLISHER_SERVICE_ACCOUNT
--- a/charts/kobe/values.yaml
+++ b/charts/kobe/values.yaml
@@ -78,6 +78,27 @@ containerSecurityContext:
 # Cluster-side defense-in-depth policies. These are opt-in because they
 # require Kubernetes features that may not be available in every cluster
 # (ValidatingAdmissionPolicy is GA in K8s 1.30+).
+# Upstream Agent Sandbox runtime (kubernetes-sigs/agent-sandbox).
+#
+# Kobe consumes SandboxTemplate / SandboxWarmPool / SandboxClaim from that
+# project. This selects how the runtime gets there.
+#
+#   disabled  (default) Sandbox features are off. Nothing is validated and no
+#             Sandbox API is served, so an upgrade never starts depending on a
+#             runtime you have not installed.
+#
+#   external  You install and own the runtime; kobe validates that the required
+#             APIs are present and version-compatible, and owns nothing.
+#
+#   managed   NOT IMPLEMENTED. It would install and RETAIN cluster-scoped CRDs,
+#             RBAC and a webhook across uninstall — Helm never deletes anything
+#             from a chart's crds/ directory, so those become permanent on any
+#             cluster that installs once. Pending approval on
+#             kunobi-ninja/kobe#72. Selecting it fails with an explicit error
+#             rather than falling back, so it cannot be switched on by accident.
+agentSandbox:
+  mode: disabled
+
 clusterPolicies:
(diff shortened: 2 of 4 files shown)


jleni:

Issue #71, opened 2026-08-10
api(sandbox): add SandboxPool, SandboxLease, and kind-aware access policy


jleni:
## Goal

Add bounded Kobe-native APIs for Agent Sandbox capacity without turning the existing Kubernetes-shaped `ClusterBackend` into a universal backend.

This is a child of #70. It defines schemas, HTTP admission, pure upstream projections, and lifecycle/provenance invariants. #73 and #74 enforce those invariants during placement and teardown; this issue must not stamp a cleanup finalizer before a controller exists to remove it.

## API contract

Add typed `SandboxPool` and `SandboxLease` CRDs plus typed HTTP request/response models.

`SandboxPool` is the operator-defined class and declares:

- desired warm capacity and TTL/provisioning limits;
- administrator-controlled containers, aggregate CPU/memory ceilings, ephemeral storage, and exposed ports;
- the claimed isolation tier, administrator-selected RuntimeClass, and readiness requirements;
- exactly one tagged placement: the management cluster, or a child `ClusterPool` reference.

The first release is ephemeral-only: no PVC, hostPath, or arbitrary volume input. This keeps #73/#74 cleanup evidence bounded; durable storage belongs to Workspace design.

`SandboxLease` declares only caller-safe intent: pool, TTL, alias, and server-derived requester metadata. It must not accept arbitrary Pod specs, RuntimeClasses, host mounts, environment/secret values, PVC specs, namespaces, or target credentials.

The upstream mapping targets the pinned Agent Sandbox v0.5.4 `v1beta1` contract selected by #72:

- `SandboxPool` -> controller-owned `SandboxTemplate` + `SandboxWarmPool`;
- `SandboxLease` -> one `extensions.agents.x-k8s.io/v1beta1/SandboxClaim`;
- Kobe expiry -> upstream `lifecycle.shutdownTime`;
- cleanup -> `shutdownPolicy: DeleteForeground` plus later Kobe verification.

Status records stable phases/conditions, expiry, placement, and non-secret target provenance including UIDs. It never contains kubeconfigs or bearer tokens. Pure helpers allow provenance to move only from absent to one exact value; #73/#74 revalidate this before target mutation.

## Policy

[...]

PR #118, opened 2026-08-12
controller(sandbox): reconcile management placement into upstream objects

emmanuelm41:
Core of #73 · **stacked on #117** (#72), which it genuinely requires

This is a true stack, not a sibling: the controller imports #72's validated-runtime module and must write the version #72 pins.

## What it does

#71 supplied the projection and the pure lifecycle transitions. This is the loop that drives them:

- each management-placement `SandboxPool` → a controller-owned `SandboxTemplate` + `SandboxWarmPool`
- each **admitted** `SandboxLease` → exactly one `SandboxClaim`, then **Ready only once that Sandbox is actually up and its shutdown is bounded without Kobe**

## Design points worth review

**Only admitted leases are placed.** A `pending` lease may exist before its quota reservation committed, so placing one would create work admission never authorised — the reason that annotation exists at all.

**`poolRef` is enforced, not decorative.** The reference carries a UID and a generation because a name is not an identity: delete and recreate `agents` with a different image, RuntimeClass or placement and the name still resolves. Quota, isolation and template were all decided against the admitted spec, so placement refuses anything else before it creates a thing.

**Names are derived, never caller-supplied.** A lease supplies a pool reference and an alias; if either reached a name here, one caller could collide with — or impersonate — another pool's template or another lease's claim. The `kobe-` prefix marks ownership, so an object without it was not created here and is never adopted.

**Pools and claims write differently, on purpose.** Pools are server-side applied, so a drifted upstream object is corrected rather than duplicated and two replicas converge instead of fighting. Claims are create-then-tolerate-409: exactly one claim per lease is the invariant, and a create that loses the race has already been satisfied by whoever won it.

**The runtime TTL starts at observed readiness, not at request.** A caller must not be charged for however long placement and provisioning took — which is precisely why the provisioning deadline is a separate bound. Readiness requires an explicit `Ready=True` condition; an absent, malformed or not-yet-populated status is "not yet", never "yes". Treating "cannot tell" as ready would hand over a lease already part-spent on a Sandbox that may never serve.

**The upstream backstop is stamped *before* the lease is marked Ready.** `shutdownTime` + `DeleteForeground` is what bounds the workload when Kobe crashes, is upgraded, or loses credentials. Marking Ready first and failing the stamp would leave a caller holding a working Sandbox whose only expiry lives in a controller that is, by assumption, not running. A failed stamp — or a missing/unparseable expiry — fails the reconcile instead.

**An already-Ready lease reuses its persisted `readyAt`.** Passing a fresh `now()` on each requeue made the transition non-idempotent: refused as a changed timestamp forever, with the backstop silently no longer re-asserted.

**An elapsed provisioning deadline releases rather than spins.** The Sandbox is shut down upstream immediately and the lease moves to `Releasing`, instead of requeueing forever against a workload that came up too late to be worth anything.

**Child-placement pools are skipped entirely**, not partially reconciled. Building their template here would create management-cluster capacity for a pool that must never serve from the management cluster. That's #74.

**Spawned only in `external` mode.** In `disabled` it would watch CRDs that may not exist and log errors forever; without #72's validation it could write objects an incompatible runtime cannot reconcile.

## Testing

Ordering is verified at the **reconcile boundary** against a mock API server, not at an extracted helper: "Ready before the backstop" and "clock started before readiness" are bugs in the *order of the writes*, and no test of a single step can see them. Both orderings were mutation-checked — inverting the write order, and forcing the readiness gate true, each fail the corresponding test.

A test also pins that the version written here is the version #72 validates — had those drifted, the operator would refuse to start against a runtime it then wrote to anyway.

## Validation

```
cargo test --workspace                                  # 1416 passed, 0 failed
```

## Scope, stated narrowly

**Not** implemented here, and remaining in #73:

- readiness gated on a real **execution canary** (`spec.readiness.canary`) in addition to the upstream `Ready` condition — the pool field exists and is projected, but nothing executes it yet, so a Sandbox whose process is wedged still reads Ready;
- driving `Releasing` → `Released`/`Expired`: nothing currently deletes the lease-owned upstream footprint or verifies its teardown, so a released lease keeps consuming quota. Verified teardown is #80's machinery applied to Sandbox subjects.

### DIFF
--- a/charts/kobe/templates/rbac.yaml
+++ b/charts/kobe/templates/rbac.yaml
@@ -48,10 +48,22 @@ rules:
     resources: ["sandboxpools"]
     verbs: ["get", "list", "watch"]
   # Sandbox callers create lease intent through the HTTP API. Lifecycle
-  # controllers, not admission, own the status subresource.
+  # controllers, not admission, own the status subresource — which needs its
+  # own rule: `/status` is a separate resource to RBAC, so a controller with
+  # `patch` on `sandboxleases` alone still 403s on every phase transition.
   - apiGroups: ["kobe.kunobi.ninja"]
-    resources: ["sandboxleases"]
+    resources: ["sandboxleases", "sandboxleases/status"]
     verbs: ["create", "get", "list", "watch", "patch", "delete"]
+  # Upstream Agent Sandbox objects, created and owned by the placement
+  # controller. Kobe never installs these CRDs (`agentSandbox.mode=external`
+  # means the operator owns the runtime); it only writes the objects it
+  # created, all of which carry a `kobe-` name and an owner reference to their
+  # Kobe parent. `delete` is what makes verified teardown possible: without it
+  # a released lease could never prove its Sandbox was gone, and its capacity
+  # would be withheld forever.
+  - apiGroups: ["extensions.agents.x-k8s.io"]
+    resources: ["sandboxtemplates", "sandboxwarmpools", "sandboxclaims"]
+    verbs: ["get", "list", "watch", "create", "update", "patch", "delete"]
   # Core resources for virtual cluster lifecycle
   - apiGroups: [""]
     resources: ["services", "configmaps", "secrets", "serviceaccounts", "pods", "pods/status", "events", "persistentvolumeclaims", "endpoints"]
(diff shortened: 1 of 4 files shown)


jleni:

PR #120, opened 2026-08-12
lifecycle(sandbox): release and expire leases against verified teardown

emmanuelm41:
Completes the lifecycle half of #73 · **stacked on #118**

## The hole this closes

The HTTP API records release intent as an annotation and deliberately never touches status — controllers own transitions. Nothing consumed that annotation. A caller who released got `204` and **kept their quota slot forever**. An expired lease leaked one too: the upstream `shutdownTime` backstop stops the Sandbox, but nothing upstream knows about Kobe's admission ledger.

## The order is the whole point

```
delete the claim  →  PROVE it absent  →  then release the reservation
```

Releasing the slot first would let it be handed to the next caller while the previous Sandbox was still running — precisely the over-subscription the ledger exists to prevent. **Absence is proven by a 404 and nothing else**; a DELETE that was accepted proves only that it was accepted.

## Three outcomes, not two

"Still there" and "cannot tell" need opposite responses, and collapsing them breaks the feature in one direction or the other:

| Observation | Meaning | Response |
|---|---|---|
| `404` | Gone | Release the slot, reach the terminal phase |
| `200` (incl. mid-deletion) | Not finished | Stay in `Releasing`, keep holding capacity |
| `401`/`403` | Evidence will never arrive | `Quarantined`, capacity withheld |

Foreground deletion takes as long as the Sandbox takes to stop. Had that window released the slot, a pool would be over-subscribed for exactly as long as teardown takes. And an operator can see and fix an under-counted pool; nobody can see a Sandbox quietly double-booked.

## Other decisions worth review

**Teardown is evaluated before the pool is resolved.** Release must work when the pool was edited or deleted outright — precisely the situations that strand capacity — so it cannot sit behind a fence a missing pool would fail.

**Expiry ends as `Expired`, a request as `Released`.** Giving capacity back and having it taken are different events to anyone reconstructing what happened to a workload.

**An unparseable `expiresAt` does not expire the lease.** Both readings of that uncertainty are wrong, but one destroys a workload someone is still using and the other leaves a visible lease an operator can act on.

**Terminal phases are never re-torn-down.** A name that has since been reused would mean acting on somebody else's footprint. `Quarantined` is terminal here on purpose: leaving it is an operator decision backed by evidence, not something a requeue may do.

**Conditions are rebuilt whole.** The status patch is a JSON merge patch, which *replaces* arrays — a patch carrying one condition would erase every other condition on the lease. `lastTransitionTime` moves only on a real flip.

## Chart

`coordination.k8s.io/leases` gains `list` and `delete`. Reservations are found by the `sandbox-lease-uid` label (a LIST, because the controller holds the lease and not the names it created) and deleted under a UID precondition. Without both verbs a lease can be torn down and still hold its slot, with nothing left to reconcile it.

## Testing

Reconcile-boundary tests against a mock API server, covering each of the three outcomes and the expiry path. Mutation-checked: reading either "still present" or "cannot look" as gone fails the corresponding test.

```
cargo test --workspace                                  # 1423 passed, 0 failed
```

## Still open in #73

- readiness gated on a real **execution canary** (`spec.readiness.canary`): the field is projected but nothing executes it, so a Sandbox whose process is wedged still reads Ready;
- GC of terminal `SandboxLease` objects. They no longer consume capacity (`consumes_capacity()` is false for `Released`/`Expired`), so this is housekeeping rather than a leak — but it is not done here.

### DIFF
--- a/charts/kobe/templates/deployment.yaml
+++ b/charts/kobe/templates/deployment.yaml
@@ -68,6 +68,11 @@ spec:
               value: {{ .Values.logLevel | quote }}
             - name: KOBE_SYNC_IMAGE
               value: {{ include "kobe.syncImage" . | quote }}
+            # `managed` is rejected by the operator at startup rather than
+            # here, so the refusal and its reason live in one place and a
+            # direct (non-Helm) deployment cannot route around it.
+            - name: AGENT_SANDBOX_MODE
+              value: {{ .Values.agentSandbox.mode | default "disabled" | quote }}
             - name: POOL_SERVICE_ACCOUNT
               value: {{ .Values.poolServiceAccount | default (include "kobe.serviceAccountName" .) | quote }}
             - name: POOL_PUBLISHER_SERVICE_ACCOUNT
--- a/charts/kobe/templates/rbac.yaml
+++ b/charts/kobe/templates/rbac.yaml
@@ -48,10 +48,22 @@ rules:
     resources: ["sandboxpools"]
     verbs: ["get", "list", "watch"]
   # Sandbox callers create lease intent through the HTTP API. Lifecycle
-  # controllers, not admission, own the status subresource.
+  # controllers, not admission, own the status subresource — which needs its
+  # own rule: `/status` is a separate resource to RBAC, so a controller with
+  # `patch` on `sandboxleases` alone still 403s on every phase transition.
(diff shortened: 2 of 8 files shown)


jleni:

PR #122, opened 2026-08-12
readiness(sandbox): gate leases on the pool's execution canary

emmanuelm41:
Closes the last functional gap in #73 · **stacked on #120**

## Why upstream `Ready` is not enough

Upstream reports `Ready` when the Sandbox's Pod is running and its probes pass. That is a statement about the **container**, not about the agent inside it. A Pod whose entrypoint crash-looped, whose weights failed to mount, or whose agent is wedged on a lock satisfies every one of those checks — and the moment Kobe believes it, the caller's **paid runtime TTL starts** on a Sandbox that cannot serve.

`spec.readiness.canary` existed on `SandboxPool` since #71 and was projected. Nothing executed it. This runs it.

## Administrator-owned, never caller-owned

`pods/exec` under the operator's credentials is a powerful thing to hold, so what bounds it matters: the argv comes from the **`SandboxPool` spec**, which only an administrator can write, and a lease selects a pool and nothing else. No caller input reaches this call. Were the argv reachable from lease intent, this would be a remote-execution primitive rather than a health check.

## Three outcomes, not two

| Result | Meaning | Response |
|---|---|---|
| exits zero | The workload itself says it works | Ready; the clock starts |
| exits non-zero | The workload says no | Not ready — retry |
| cannot be run | No evidence either way | Not ready — retry |

"Failed" and "could not run" both stop readiness, but they are not the same fact, and neither is a pass. Either way **nothing is written**: no clock, no shutdown time, no `Ready`. The provisioning deadline (already enforced in #118) bounds how long that repeats, so a Sandbox that never passes ends as an expiry rather than as a lease that hangs — no new stuck state is introduced.

## Decisions worth review

**The Pod is found through upstream's own fields.** `claim.status.sandbox.name` → the `Sandbox` object → `status.selector` → the Pod. Never a naming convention: a convention upstream later changes would silently start selecting a Pod belonging to **another tenant**. An ambiguous selector is an error rather than a coin flip, and a terminating Pod is excluded — it reports `Running` for the whole of its grace period.

**The exec is bounded by the pool's own timeout.** Without it a wedged agent — precisely the failure this catches — would hang the reconcile and take a controller worker slot with it, turning a detectable fault into an outage. A malformed timeout refuses to run the canary at all rather than running it unbounded.

**A pass is recorded on the lease and never re-run.** Repeating an administrator's command inside a live tenant workload on every requeue is a recurring side effect, not a health check. The record is durable (a `ReadinessCanary` condition) rather than in-memory, so a controller that restarted between the pass and the `Ready` write does not execute it a second time.

**`sandboxes.agents.x-k8s.io` joins the required-CRD set**, at its own group version — it is in the *core* Agent Sandbox group, not the `extensions.` one. Its absence is the one that fails **silently**: it is read to find the Pod, so a runtime without it produces leases that never pass readiness rather than an error anyone can act on. `crd_is_compatible` now takes the expected API version per CRD instead of assuming one group.

## Chart

- `agents.x-k8s.io/sandboxes`: read-only. Kobe creates Sandboxes only indirectly, through claims.
- `pods/exec`: `create`, for the canary.

## Testing

```
cargo test --workspace                                  # 1430 passed, 0 failed
```

Mutation-checked: treating an inconclusive canary as a pass, and treating any non-failure as a pass, each fail their tests. The reconcile-boundary test drives a claim upstream calls `Ready` whose Sandbox cannot be reached, and asserts that neither the shutdown time nor the lease status is written.

> The canary itself execs over a websocket, which the HTTP mock cannot serve. Tests covering what happens *after* readiness record the canary pass the way a previous reconcile would have — which is also the production path after any controller restart. What is **not** covered by unit tests is the live exec round-trip; that belongs to #76's conformance suite.

### DIFF
--- a/charts/kobe/templates/rbac.yaml
+++ b/charts/kobe/templates/rbac.yaml
@@ -64,6 +64,23 @@ rules:
   - apiGroups: ["extensions.agents.x-k8s.io"]
     resources: ["sandboxtemplates", "sandboxwarmpools", "sandboxclaims"]
     verbs: ["get", "list", "watch", "create", "update", "patch", "delete"]
+  # The upstream `Sandbox` object itself, in the CORE Agent Sandbox group.
+  # Read-only: Kobe creates Sandboxes only indirectly, through claims. It is
+  # read to find the Pod behind a claim, via upstream's own `status.selector`.
+  - apiGroups: ["agents.x-k8s.io"]
+    resources: ["sandboxes"]
+    verbs: ["get", "list", "watch"]
+  # The readiness canary runs an administrator-declared command inside the
+  # Sandbox. Upstream `Ready` describes the container, not the agent in it —
+  # a crash-looped or wedged agent satisfies it, and believing it starts a
+  # paid runtime TTL on a Sandbox that cannot serve.
+  #
+  # `pods/exec` is a powerful verb, so what bounds it matters: the argv comes
+  # from the `SandboxPool` spec, which only an administrator can write, and a
+  # lease selects a pool and nothing else. No caller input reaches this.
+  - apiGroups: [""]
+    resources: ["pods/exec"]
+    verbs: ["create"]
   # Core resources for virtual cluster lifecycle
   - apiGroups: [""]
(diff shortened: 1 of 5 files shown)


jleni:

PR #124, opened 2026-08-12
ci: run the E2E harness lint/tests in their own job

jleni:
## Why

A TypeScript lint took a production hotfix hostage today.

`checks` (the only required Rust check) installed bun via mise so it could lint and test `hack/*.ts`. mise fetches bun from `release-assets.githubusercontent.com`, and the self-hosted `kunobi-runners` egress refuses that host outright — **instant** HTTP/2 `refused stream`, ~70ms in, not a timeout — in stretches lasting the better part of an hour:

```
  client error (SendRequest): http2 error: stream error received: refused stream
```

Every branch went red simultaneously (`fix/teardown-fence-deadlock`, `chore/bump-github-actions`, `feat/agent-sandbox-external-mode`), so nothing could merge — including #119, the fix for the incident that had the int-pro pool exhausted and CI unable to lease a cluster.

The asset is not the problem: GitHub-hosted egress pulls it fine, and so does a laptop (`curl -L` → 200). This is specific to the self-hosted fleet's egress.

## What

- The harness lint/tests move to their own `e2e-harness` job on `ubuntu-latest`. Rust checks stop depending on a tool host they never needed.
- Retries stay (three bounded attempts) — the refusal flaps, which is why #121's run passed on attempt 2 while #119's failed all three.
- The mise cache stays on; tools are pinned in `mise.toml`, so a warm run needs no network for them at all.

## Trade-off worth naming

`e2e-harness` is not in branch protection's required set, so a harness failure now reports without blocking. That is intended — the lint guards `hack/*.ts`, not the operator. Add it to the required set later if you'd rather it gate merges, ideally once the runner egress issue is understood.

The underlying egress refusal is still worth an infra ticket; this only removes it from the critical path.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -67,37 +67,6 @@ jobs:
     steps:
       - uses: actions/checkout@v7
 
-      # `cache: false` meant every single run re-downloaded bun, so one bad
-      # minute on the tool host took the whole required check down — as it did
-      # on 2026-08-12, when `mise install bun` failed instantly with an HTTP/2
-      # `refused stream` on every branch for the better part of an hour and
-      # nothing could merge. The tool is version-pinned in mise.toml, so the
-      # cache key is exact and a hit is the normal path.
-      - name: Set up mise
-        uses: jdx/mise-action@v4
-        with:
-          install: false
-          cache: true
-
-      # Retried separately rather than via `install_args`: a transient refusal
-      # on the download is not a reason to fail a Rust check, and the action
-      # itself has no retry. Bounded — three attempts, then the failure is
-      # real and the log shows all of them.
-      - name: Install bun for the E2E harness tests
-        run: |
-          for attempt in 1 2 3; do
-            if mise install bun; then exit 0; fi
-            echo "::warning::bun install attempt ${attempt} failed; retrying"
-            sleep $((attempt * 15))
-          done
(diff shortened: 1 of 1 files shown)


jleni:
Closing: we're keeping the harness on the self-hosted fleet and fixing the download at the source instead of moving work to GitHub-hosted runners.

Investigation summary for the record — the refusal was **not** a config break and not the GitHub API rate limiter:

- No proxy is configured on the `kunobi-runners` AutoscalingRunnerSet, and no CiliumNetworkPolicy applies to the `remotedev` namespace, so cilium-envoy is not in the path. The `REFUSED_STREAM` comes from the far end.
- The dind path is not misconfigured either: `docker0` is MTU 1500 inside a pod `eth0` of 9050, and downloads succeed through both the pod netns and a container on dind.
- Re-tested after the window closed: 5/5 `mise install bun@1.3.14` inside `rust:1.94-bookworm` on dind, and 10/10 parallel asset fetches — all clean.

Consistent explanation: connection-level throttling of our shared egress IP (all runners NAT out one address, up to 21 concurrent across both orgs) surfacing as an HTTP/2 GOAWAY/REFUSED_STREAM, which mise's client reports as a hard error while curl transparently retries on a fresh connection. That is why curl never failed in the same environment while mise failed on every branch at once.

The retry + mise cache from #121 stay; the durable fix is to stop fetching pinned tools from GitHub's edge on every run.

PR #125, opened 2026-08-12
placement(sandbox): compose child-cluster Sandboxes through an internal ClusterLease

emmanuelm41:
Core of #74 · **stacked on #122**

A child-placement `SandboxPool` says: do not run this pool's Sandboxes next to everybody else's. Kobe acquires one exclusive cluster from the named `ClusterPool` and places the caller's Sandbox inside it.

## ⚠️ Scope narrowed, deliberately — please read this first

#74 as filed has Kobe **bootstrap the Agent Sandbox runtime into the child cluster** and verify it with a create/delete canary. Installing cluster-scoped CRDs, RBAC and a webhook with cluster-admin is one of the three effects **paused pending approval on #72**, and #72 shipped `external` mode only.

Child placement follows the same rule: the child cluster must **already serve** a compatible runtime — through the `ClusterPool`'s own `bootstraps` or `addons` — and Kobe *validates* it with the same check #72 applies to the management cluster. A child without one is refused with a legible error telling the operator to configure it.

This is a real narrowing, and it is a governance call rather than a technical one: a controller must not perform under a different issue number the effect it was told to hold. If you want the bootstrap half, that is the approval on #72, and it can land on top of this without rework.

## Composition

**Every refusal happens before a cluster is allocated.** An abandoned child composition costs a whole cluster's capacity, so eligibility, lifetime and identity are all settled first:

- **Backend.** Only a receipt-capable k3s pool is admitted; everything else is `Unsupported`, never silently weaker. An exclusive tenant cluster whose destruction nobody can prove defeats the point of giving them one.
- **Lifetime.** The internal lease covers provisioning **+ the full runtime TTL + a drain grace**, and a `ClusterPool` whose ceiling cannot reach that is refused. A cluster that expired mid-lease would take the running Sandbox with it; one that expired mid-teardown would take the evidence that it was cleaned up — which is worse, because capacity would be reused with nothing able to prove it was safe.
- **Identity.** The internal lease is named from the Sandbox lease, so a restarted controller resumes the same cluster rather than allocating a second. Adoption is fenced on ownership: a same-named lease this Sandbox does not own is somebody else's cluster.

**It is owner-referenced to the `SandboxLease`.** That is what makes "cancelling a pending outer lease releases any partial internal allocation" true without a compensating code path — and a compensating path is exactly what fails to run when the controller is the thing that crashed.

**Its requester is Kobe, not the caller.** Attributing it to the tenant would surface the internal cluster in their own listings and read as theirs to release.

**Provenance is written before the cluster is used.** Teardown must be able to name the exact instance it has to prove absent, and provenance written only after a successful placement is missing in precisely the case that matters — a crash partway through.

**The child kubeconfig never leaves memory.** Not into status, not into an API response, not into a log line. The kubeconfig-read error is swallowed rather than propagated, because its context can carry the secret's own contents.

**Upstream objects in a child carry no owner reference.** An owner reference names an object in the *same* cluster; a reference to a management-cluster `SandboxPool` would name nothing there and Kubernetes GC would delete the template on sight. A child's objects are bounded by the cluster's own lifetime instead.

## Teardown — where the two placements most dangerously diverge

A child lease is released by **destroying its cluster**, not by deleting a claim. The claim lives in the child, so deleting it against the management cluster returns 404 — and a 404 is precisely what that path treats as **proof of absence**. The quota slot would be handed to the next caller while the previous tenant's Sandbox was still running in a cluster nobody had touched. This is asserted directly rather than left to inspection.

It acts on the **recorded UID**. A later composition reusing the name is not this lease's cluster: destroying it would kill a running tenant, and treating its presence as "ours is still up" would strand this lease forever. Neither — the recorded cluster is provably absent, so the release completes.

Not being able to tell **quarantines**.

Both placements share one release tail, so their release semantics cannot drift apart — the equivalence #76 sets out to prove, built in rather than asserted.

## A leak this work would otherwise have opened

The HTTP API returned `status.target` verbatim. For a child lease that carries the internal `ClusterLease` and `ClusterInstance` **name and UID** — the exact identifiers every cluster-lease endpoint is keyed on. #74 requires the composition be *undiscoverable*, not merely unusable; "no authority" and "no authority, but knows precisely what to ask for" are not the same posture.

Stripped at the response boundary rather than never recorded: teardown must prove that exact instance UID gone, and evidence a controller cannot see is evidence that does not exist. The test asserts on the **serialized** form, so a field renamed into the wire format later cannot pass a struct-level check and still leak.

## Testing

```
cargo test --workspace                                  # 1446 passed, 0 failed
```

Mutation-checked: letting a child lease fall through to the management teardown, matching the recorded cluster by name instead of UID, accepting any receipt regardless of which instance it describes, and completing when no receipt exists at all — each fails its test.

## Against #74's acceptance criteria

| Criterion | Here |
|---|---|
| Only a receipt-capable k3s pool admitted; others `Unsupported` before binding | ✅ |
| Internal cluster cannot expire before the outer cleanup window | ✅ |
| Cancelling a pending outer lease releases partial internal allocation | ✅ (owner reference) |
| Caller can neither discover nor operate the backing `ClusterLease` | ✅ (incl. the API fix) |
| No kubeconfig or admin token in status, responses, logs or events | ✅ |
| Equivalent release/quarantine semantics across placements | ✅ (shared tail) |
| Stale/name-reused target provenance fails closed | ✅ |
| Reconciliation resumes safely after restart at every transition | ⚠️ partial — resumption is covered at composition and teardown; the full matrix belongs to #76 |
| Cleanup terminal only after a receipt proves the original instance absent | ✅ |

## Completion requires the receipt

The internal lease is **released, never deleted**. Deleting the object takes its status — and the teardown receipt on it — with it, at exactly the moment the evidence matters; #80 put the receipt on the lease rather than the instance for precisely this reason. The object is collected later by its owner reference.

A release completes only against a receipt that is complete, self-consistent, and about **this** lease's instance:

- outcome `Verified`, an attempt that finished, and checks that agree with the stated outcome — the receipt is not trusted over its own evidence;
- a schema version this build understands;
- the instance UID recorded at composition time. A child cluster can be destroyed and replaced under the same lease name, and a receipt proving the *replacement* gone says nothing about the instance this Sandbox ran on.

Everything else quarantines: a receipt about a different instance, a child whose own teardown quarantined, or a recorded lease that is gone or whose name now belongs to somebody else's composition. Present-but-wrong is the dangerous case, because it is the one a laxer check accepts.

## Not in this PR

**Bootstrapping the child runtime.** See the scope note above; it is #72's approval.

### DIFF
--- a/src/api/sandbox.rs
+++ b/src/api/sandbox.rs
@@ -640,8 +640,32 @@ fn sandbox_lease_response(
         ready_at: status.ready_at,
         expires_at: status.expires_at,
         placement: status.placement,
-        target: status.target,
+        target: status.target.map(caller_visible_provenance),
         conditions: status.conditions,
+    }
+}
+
+/// Strip the internal composition from provenance before it reaches a caller.
+///
+/// `SandboxTargetProvenance` is the controller's restart-safe record of exactly
+/// what it built, and for a child-placement lease that includes the internal
+/// `ClusterLease` and `ClusterInstance` Kobe acquired to host the Sandbox.
+///
+/// The caller asked for a Sandbox. The cluster underneath it is Kobe's
+/// implementation of the pool, not a capability they hold — they cannot
+/// connect to it, extend it or release it, and #74 requires that they cannot
+/// *discover* it either. Returning its name and UID would hand them the exact
+/// identifiers every cluster-lease endpoint is keyed on, turning "no
+/// authority" into "no authority, but knows precisely what to ask for".
+///
+/// Stripped at the response boundary rather than never recorded: teardown must
+/// be able to prove that exact instance UID gone, and evidence a controller
+/// cannot see is evidence that does not exist.
+fn caller_visible_provenance(target: SandboxTargetProvenance) -> SandboxTargetProvenance {
+    SandboxTargetProvenance {
+        child_cluster_lease: None,
+        child_cluster_instance: None,
+        ..target
     }
 }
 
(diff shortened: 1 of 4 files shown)


jleni:

Issue #83, opened 2026-08-10
access(sandbox): add bounded TTY attach and declared-port forwarding


jleni:
## Goal

Add bounded TTY attach/exec and declared-port forwarding over the restricted Sandbox access substrate, with immediate lease-scoped stream revocation.

This is a child of #75. Interactive transport is human-facing convenience; non-TTY Execution remains the agent contract.

## Contract

- Support the exact Kubernetes upgrade protocols required for exec/attach/port-forward without exposing a general API path.
- Resolve only the owned, Ready, unexpired lease's exact Pod UID, approved container, and pool-declared remote port.
- Enforce operation timeout, idle timeout, maximum duration, byte limits, and per-principal/per-lease concurrency.
- Support TTY resize and bounded signal forwarding where the selected runner/container contract supports them.
- Strip authorization, impersonation, forwarding, and hop-by-hop headers; construct target requests from allowlisted fields only.
- Register every live stream against its lease UID on the serving replica. Release, expiry, quarantine, target loss, or UID change cancels the stream before destructive cleanup.
- Port forwarding binds locally by default; public service exposure is out of scope.

## Acceptance criteria

- [ ] Direct and child placements expose identical TTY and port-forward behavior.
- [ ] Another identity receives 404 and cannot operate or discover a stream.
- [ ] Callers cannot choose another namespace, Sandbox, Pod, container, undeclared port, path, protocol, or header.
- [ ] Terminal resize/signals work where declared and fail explicitly where unsupported.
- [ ] Idle/duration/byte/concurrency limits terminate streams deterministically.
- [ ] Release and expiry actively close existing streams on every API replica before target cleanup proceeds.
- [ ] Cached contexts cannot outlive lease expiry, revocation, or any UID/provenance change.
- [ ] Scoped credentials and stream content never appear in responses beyond the intended stream, CRDs, events, audit logs, or test artifacts.
- [ ] Existing ClusterLease exec/attach/port-forward compatibility remains unchanged.

## Non-goals

- Non-TTY execution persistence/idempotency
- General Kubernetes proxying
- Public ingress or Agent Sandbox Router
- Shared interactive sessions
- SSH, iroh, or Herdr

## Dependencies

- Depends on: the restricted access and non-TTY execution children of #75
- Blocks: the CLI child and #76
- Parent: #75

jleni:
[...]

Issue #75, opened 2026-08-10
access(sandbox): add agent-native Execution API, restricted broker, and CLI


jleni:
## Goal

Add an agent-native execution contract for `SandboxLease` that requires neither an interactive terminal nor Kubernetes knowledge/credentials. Human attach is a convenience over the same restricted path.

Depends on #71 and the exact target resolution delivered by #73 / #74. This is a child of #70.

## Delivery breakdown

This issue is the parent contract for four formal sub-issues, delivered in dependency order:

- [ ] #81 — exact target resolution, scoped credentials, operation allowlists, revocation, and audit baseline
- [ ] #82 — durable idempotent non-TTY execution and result records
- [ ] #83 — bounded TTY attach and declared-port forwarding
- [ ] #84 — CLI commands and `run` lifecycle composition

#82 depends on #81. #83 depends on #81 and #82. #84 depends on #82 and #83. This parent closes only after all four children satisfy the shared contract below.

## Execution API

Expose a lease-scoped API equivalent to:

```text
POST   /v1/sandbox-leases/{lease}/executions
GET    /v1/sandbox-leases/{lease}/executions/{execution}
GET    /v1/sandbox-leases/{lease}/executions/{execution}/logs
DELETE /v1/sandbox-leases/{lease}/executions/{execution}
```

An execution request uses structured fields: `argv[]`, optional `cwd`, `tty` (false by default), bounded `timeout`, detach/wait mode, and an idempotency key. Kobe never adds an implicit shell.

Results expose a stable execution ID, state, timestamps, separate stdout/stderr streams, and exact remote exit status. JSON/NDJSON modes must be easy for an LLM agent to parse. A detached execution survives client disconnect and can be polled, tailed, or cancelled, but never outlives its Sandbox lease/target Pod.

## Restricted broker

Resolve every operation fail-closed through:

```text
authenticated principal
  → owned, Ready, unexpired SandboxLease UID
  → exact placement and target UIDs
  → upstream Sandbox
  → current Pod and approved container/ports
```

- Do not reuse the unrestricted cluster connect proxy as a public Sandbox surface.
- Use a lease-specific, namespace-scoped target identity with short-lived TokenRequest credentials (or an equivalently least-privileged internal mechanism).
- Permit only the exact Kobe-owned Pod/container for exec/attach/logs and only pool-declared ports for forwarding.
- Canonicalize/allowlist methods, paths, subresources, stream protocols, headers, and ports.
[...]

PR #126, opened 2026-08-12
access(sandbox): resolve operations to one exact owned target

emmanuelm41:
Substrate of #81 · **stacked on #125**

Every Sandbox operation — exec, logs, attach, port-forward — has to answer one question first: **which Pod, in which cluster, for which caller.** This is that substrate, landed together with its first consumer so the resolver is exercised rather than merely written.

## "Never a default" is the whole design

A resolver that falls back — to the current namespace, the first container, the only Pod it happened to find — turns a stale or ambiguous request into a **successful** one against the wrong workload. For a Sandbox that means a caller's command running inside somebody else's agent.

So every step yields the exact object placement recorded, or denies:

```
authenticated principal
  -> owned, Ready, unexpired SandboxLease UID
  -> immutable direct/child placement provenance
  -> exact Claim/Sandbox/Pod UIDs
  -> approved container
```

## Decisions worth review

**Absent and unowned are the same answer.** A 403 confirms the lease exists, and that difference is enough to enumerate every lease in the namespace by guessing names — which is precisely what ownership checking exists to prevent. No denial variant maps to 403, the 404 carries no body, and a test asserts both.

**The audit reason code is finer-grained than the caller-facing status.** An operator has to tell "expired" from "never placed" when somebody reports that access stopped working; the caller must not be able to. The audit line carries principal, lease, operation, target UID and outcome — never a credential, a command, or the workload's output.

**`Releasing` denies even though the Pod may still be up.** Teardown has started. Letting a new stream attach to a workload being destroyed gives a stream that dies mid-flight at best, and interferes with cleanup at worst.

**An unreadable or absent expiry denies.** Treating it as "not expired" grants access on the strength of a value nobody can read — and bounding what the caller may keep doing is the entire job of the expiry.

**Missing provenance denies rather than triggering a lookup.** Looking the Pod up at access time finds whatever holds the name *now*; after a recycle that is another tenant's workload. A reference with an **empty UID** also denies: two empty UIDs compare equal, so a later same-named Pod would satisfy the check meant to exclude it.

**Only the pool's own container is addressable.** Anything else in the Pod is the administrator's — a sidecar, an injected proxy — and executing there runs the caller's command with that component's identity and mounts rather than their own.

**A pool recreated under the same name denies** rather than supplying a new allowlist. Its template is a different administrator's decision about what should be reachable, and silently applying it changes what an already-admitted caller can do.

## Placement now records what the resolver consumes

The exact `SandboxClaim` / `Sandbox` / `Pod` identities, written when those objects have just been observed, and **monotonically** — provenance that could be cleared would let a teardown lose the identity it has to prove absent.

## First consumer: a bounded log tail

Deliberately the smallest operation that exercises the whole chain without needing a stream protocol.

- The tail is **clamped**, not rejected — refusing an over-large request teaches callers to retry in a loop, which costs more than serving the cap once.
- `follow` is never set: this returns a bounded body, and a followed stream here would hold a connection open with no revocation path.
- Unknown query fields are **refused rather than ignored**. `follow`, `previous`, `sinceSeconds` and `limitBytes` are all real `LogParams` options; silently dropping one a caller sent means they believe they set a bound that was never applied.
- The Pod's UID is re-checked against provenance before its output is returned.
- Child placement is refused explicitly rather than read from the management cluster, which would return another workload's output or a misleading 404.

## Chart

`pods/log` (`get`) — its own resource to RBAC, and worth stating explicitly: this is the verb that lets the operator return a workload's own output to its lease holder.

## Testing

```
cargo test --workspace                                  # 1455 passed, 0 failed
```

Mutation-checked: honouring a caller-named container, treating an unparseable expiry as valid, and accepting an empty UID each fail their tests.

## Against #81's acceptance criteria

| Criterion | Here |
|---|---|
| Management and child placement resolve through the same typed interface | ✅ |
| Foreign identities get 404, indistinguishable from absent | ✅ |
| Pending/Releasing/Released/Expired/Quarantined/malformed-expiry/missing-target/stale-UID deny before minting | ✅ |
| Namespace, Claim, Sandbox, Pod, container, pool substitution fail closed | ✅ |
| Only approved containers resolve | ✅ |
| Only pool-declared ports resolve | ⚠️ the allowlist is carried on the target; the resolver method lands with #83's port-forward, which consumes it |
| Audit without credentials, stdin, env, stdout or stderr | ✅ |
| Scoped credentials cannot read Secrets, mutate RBAC, create Pods… | ❌ **not yet** — see below |
| Release/expiry cancels already-upgraded streams on every replica | ❌ **not yet** — no streams exist until #83 |
| Existing ClusterLease connect behaviour unchanged | ✅ untouched |

## Not in this PR

**Per-lease credential minting** (ServiceAccount / Role / RoleBinding / TokenRequest in the target namespace). Today the operator's own client reaches the Pod, bounded by the resolver rather than by a scoped token. The resolver is the precondition for minting — you cannot scope a credential to a target you have not resolved — and it is the larger half; minting is the next layer.

**Stream revocation on release.** There are no upgraded streams yet: the log read returns a bounded body. This becomes real with #83 and lands with it.

Both are called out rather than glossed, and neither weakens what is here — the resolver denies before any of it is reached.

### DIFF
--- a/charts/kobe/templates/rbac.yaml
+++ b/charts/kobe/templates/rbac.yaml
@@ -81,6 +81,12 @@ rules:
   - apiGroups: [""]
     resources: ["pods/exec"]
     verbs: ["create"]
+  # Sandbox log reads (#81). Separate from `pods` above because `pods/log` is
+  # its own resource to RBAC — and worth stating explicitly: this is the verb
+  # that lets the operator return a workload's own output to its lease holder.
+  - apiGroups: [""]
+    resources: ["pods/log"]
+    verbs: ["get"]
   # Core resources for virtual cluster lifecycle
   - apiGroups: [""]
     resources: ["services", "configmaps", "secrets", "serviceaccounts", "pods", "pods/status", "events", "persistentvolumeclaims", "endpoints"]
--- a/src/api/mod.rs
+++ b/src/api/mod.rs
@@ -3,4 +3,5 @@ pub mod connect;
 pub mod policy;
 pub mod routes;
 pub mod sandbox;
+pub mod sandbox_access;
 pub mod upgrade;
--- a/src/api/sandbox.rs
+++ b/src/api/sandbox.rs
@@ -6,7 +6,7 @@
 //! or expose upstream Agent Sandbox objects. Placement controllers in #73/#74
 //! own those transitions.
 
-use axum::extract::{Path, State};
+use axum::extract::{Path, Query, State};
 use axum::http::StatusCode;
 use axum::response::{IntoResponse, Response};
 use axum::routing::{get, post};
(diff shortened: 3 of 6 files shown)


jleni:

Issue #82, opened 2026-08-10
execution(sandbox): persist idempotent non-TTY commands and results


jleni:
## Goal

Provide durable-for-the-lease, idempotent non-TTY command execution with stable status, separate stdout/stderr, exact exit status, cancellation, and honest restart/disconnect behavior.

This is a child of #75 and consumes its restricted target/credential substrate.

Plain Kubernetes `exec` is not sufficient: it cannot provide crash-safe idempotency, reconnectable detached output, or a trustworthy unknown-outcome state. Ship and qualify a Kobe runner/supervisor contract inside an administrator-selected Sandbox container; the first supported pool profile may use a first-party runner image.

## Execution contract

- Structured request: `argv[]`, optional `cwd`, bounded timeout, wait/detach mode, and required idempotency key. Kobe never inserts a shell.
- Reserve the idempotency key durably in the target before spawning. The same key and request digest returns the original execution; the same key with another digest conflicts.
- The runner owns process group creation, timeout, cancellation, stdout/stderr capture, exit status, and bounded retention until lease teardown.
- Kobe persists a typed execution record containing IDs, request digest, state, timestamps, bounded reason, target UID, and exit status. Command arguments and output do not enter Kubernetes status/events/logs.
- States include queued/running/succeeded/failed/cancelled/timed-out/unknown. A control-plane restart or lost acknowledgement must never trigger a duplicate spawn; unverifiable outcomes become `Unknown`.
- An execution never outlives its exact SandboxLease/Pod UID. Release/expiry cancels it before cleanup.
- JSON and NDJSON responses remain stable and keep stdout and stderr distinct.

## API

```text
POST   /v1/sandbox-leases/{lease}/executions
GET    /v1/sandbox-leases/{lease}/executions/{execution}
GET    /v1/sandbox-leases/{lease}/executions/{execution}/logs
DELETE /v1/sandbox-leases/{lease}/executions/{execution}
```

## Acceptance criteria

- [ ] Non-TTY wait mode returns distinct stdout/stderr and the exact remote exit code.
- [ ] Detached execution survives client disconnect and supports poll, log tail, and cancel.
- [ ] Reusing one idempotency key cannot start a duplicate command across retries or API/controller restarts.
- [ ] Conflicting request content for one idempotency key is rejected.
- [ ] Failure injection covers crash before spawn, after spawn before acknowledgement, and after acknowledgement before Kobe status persistence.
[...]

PR #127, opened 2026-08-12
access(sandbox): mint per-lease single-Pod credentials, and add bounded exec

emmanuelm41:
Completes #81 · **stacked on #126**

The resolver answers **which** Pod. This answers **with what authority** Kobe touches it.

## Why the operator's own client is not good enough

It is cluster-admin-adjacent. A bug anywhere in the request path — a mis-parsed name, a header that survived, a proxy that followed a redirect — is then a bug with the operator's authority behind it.

Kobe now mints a ServiceAccount whose Role names **exactly one Pod**, requests a short-lived token for it, and reaches the tenant's Sandbox as that identity. A credential that *cannot* name a second Pod turns those bugs from privilege escalation into a 403.

This is defence in depth, not the primary control — the resolver still denies before anything is minted. The two fail differently, which is the point of having both.

## The boundary

**`resourceNames` is the entire boundary.** An RBAC rule without it grants the verb over every Pod in the namespace — in a management cluster, that is every tenant's Sandbox at once. A test asserts every rule carries it.

**Each operation gets only its own subresources.** Reading logs does not carry the verb that runs commands. If these ever collapsed into one rule set, a `logs` request would be minting an exec-capable token.

**Nothing reaches Secrets, RBAC, Nodes, namespaces, or `list`/`watch`/`delete`.** Enumerated in a test rather than left implicit: this is the list #81 commits to, and a rule set that grew one should fail here rather than in an audit.

**Credential objects are named by lease UID, not lease name.** Names recycle. A Role left behind by a previous lease and adopted by its same-named successor would scope the new caller's credential to the **old** caller's Pod — a cross-tenant grant produced entirely by naming.

**Tokens live ten minutes and are never persisted.** A token in a CRD, a response, an event or a log is one somebody can replay after the lease it belonged to is gone.

**The scoped client keeps the operator's endpoint and trust anchors but drops every other authorisation.** An inherited client certificate would out-rank the bearer token and silently restore exactly what this removes.

Roles are server-side applied so a drifted one is corrected rather than trusted — drift matters more than usual here, because a Role edited to drop its `resourceNames` is a namespace-wide grant that looks, by name, identical to the scoped one.

## Bounded exec

The request/response surface #81 describes. Not #82: no persistence, no idempotency.

- **No shell.** argv is executed directly, so quoting is never the security boundary.
- **No stdin, no TTY.** Both need a stream protocol with its own revocation story — #83.
- **Output is capped, and the truncation is reported.** The caller chooses what runs, so they choose how much it prints; `cat /dev/urandom` from inside a sandbox is otherwise a way to make the operator buffer unbounded memory on request. Reporting matters as much as capping: a caller parsing partial output as complete turns a cap into a wrong answer.
- **A wall-clock bound**, because `sleep infinity` would otherwise hold an API worker permanently.
- **An unreported exit status is not success.** "We could not tell" must not read as "it worked".
- **Unknown request fields are refused, not ignored** — `tty`, `stdin`, `pod`, `namespace` are all things a caller might try to send, and `command` has no default.

## A correction to #126

Two tests were silently deleted by a later edit in that PR — the log tail bound and the log query-field rejection — while its body claimed both were covered. The behaviours were correct throughout; only the coverage was lost. Both are restored here.

## Chart

`serviceaccounts/token` (`create`). Bounded by RBAC's own escalation rule: a Role may only be created granting permissions the creator already holds, so this cannot manufacture authority the operator lacks — only hand out less.

## Testing

```
cargo test --workspace                                  # 1465 passed, 0 failed
```

Mutation-checked: dropping the `resourceNames` pin, and letting logs borrow exec's authority, each fail their tests.

## #81 acceptance criteria now met

| Criterion | |
|---|---|
| Scoped credentials cannot read Secrets, mutate RBAC, create Pods, access Nodes, or cross namespaces | ✅ |
| Credentials short-lived, never in CRDs, responses, events, logs or test artifacts | ✅ |
| Lease-specific ServiceAccount/Role/RoleBinding + TokenRequest in the exact target namespace | ✅ |
| Only the exact Kobe-owned Pod/container subresources the operation needs | ✅ |

Still open, and landing with the PRs that create the conditions for them:

- **Stream revocation on release** — there are still no upgraded streams; exec here is request/response. This becomes real with #83 and lands with it.
- **Port-forward's port allowlist** — the declared set is carried on the target; the resolver method lands with #83, which consumes it.
- **Child placement** for both operations — refused explicitly rather than read from the wrong cluster.

### DIFF
--- a/charts/kobe/templates/rbac.yaml
+++ b/charts/kobe/templates/rbac.yaml
@@ -87,6 +87,17 @@ rules:
   - apiGroups: [""]
     resources: ["pods/log"]
     verbs: ["get"]
+  # Per-lease scoped credentials (#81). Kobe mints a ServiceAccount whose Role
+  # names exactly one Pod, then a short-lived token for it, and reaches the
+  # tenant's Sandbox as THAT identity rather than as itself.
+  #
+  # `serviceaccounts/token` is the mint. It is bounded by RBAC's own escalation
+  # rule: a Role may only be created granting permissions the creator already
+  # holds, so this cannot manufacture authority the operator lacks — it can
+  # only hand out less.
+  - apiGroups: [""]
+    resources: ["serviceaccounts/token"]
+    verbs: ["create"]
   # Core resources for virtual cluster lifecycle
   - apiGroups: [""]
     resources: ["services", "configmaps", "secrets", "serviceaccounts", "pods", "pods/status", "events", "persistentvolumeclaims", "endpoints"]
--- a/src/api/mod.rs
+++ b/src/api/mod.rs
@@ -4,4 +4,5 @@ pub mod policy;
 pub mod routes;
 pub mod sandbox;
 pub mod sandbox_access;
+pub mod sandbox_credentials;
 pub mod upgrade;
--- a/src/api/sandbox.rs
+++ b/src/api/sandbox.rs
@@ -65,6 +65,92 @@ pub fn routes<B: ClusterBackend + Clone + 'static>() -> Router<AppState<B>> {
             get(get_sandbox_lease::<B>).delete(release_sandbox_lease::<B>),
         )
         .route("/v1/sandbox-leases/{id}/logs", get(sandbox_logs::<B>))
+        .route("/v1/sandbox-leases/{id}/exec", post(sandbox_exec::<B>))
+}
+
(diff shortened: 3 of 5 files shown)


jleni:

PR #128, opened 2026-08-12
access(sandbox): revoke in-flight operations when a lease stops permitting access

emmanuelm41:
Revocation half of #83 · **stacked on #127**

## The problem a token expiry does not solve

Kubernetes authenticates an upgraded connection **once**, at upgrade time. After that it stays authenticated for as long as it stays open. A ten-minute token does not close an hour-old session, and neither does deleting the ServiceAccount behind it.

So a lease that is released, expires, or quarantines has to **actively** cancel its operations. Otherwise the caller keeps working inside a workload whose lease has ended — and teardown proceeds underneath them, which means the "verified absent" a #80 receipt reports was verified while somebody was still using it.

## Design points

**Only `Ready` keeps an operation alive.** `Releasing` is the case that matters and the least obvious: the Pod may still be up, so a check that only looked at *terminal* phases would leave a caller working inside something whose absence cleanup was busy proving.

**Registration is keyed by lease UID, never by name.** A recreated same-named lease is a different lease. Keyed by name, revoking the old one would kill the *new* caller's session, and revoking the new one would leave the *old* caller's session open right through teardown.

**Registrations are keyed by an id, not held in a list.** Cancellation tokens have no identity to compare on — two streams of the same lease are indistinguishable by value — so removing "one of them" would deregister somebody else's.

**Deregistration happens in `Drop`.** A panicking or early-returning handler cannot leak a registration that a later revocation would then report cancelling, long after the socket was gone.

**Per-lease concurrency is bounded.** A lease is one Sandbox with one container; concurrency past a handful is not a caller working faster, it is a caller holding the operator's connections. The bound is **per-replica** because those connections are exactly what it protects — a global count would need coordination that could fail open — and **per-lease**, so one busy caller cannot lock everybody else out.

## Every replica, not a leader

A replica can only cancel connections it is holding itself: the socket lives in one process. A leader-elected revoker would leave live streams on every *other* replica with nobody watching them, which is worse than no revoker at all because it would look like the feature was working.

So each API replica runs its own watch over the same lease objects. There is no coordination and there must not be.

## First registered operation

Exec. Releasing the lease now cancels a running command with `410 Gone` rather than letting it finish inside a workload being destroyed. The concurrency limit returns `429`, audited by cause like every other denial.

## Testing

```
cargo test --workspace                                  # 1472 passed, 0 failed
```

Mutation-checked: treating `Releasing` as still permitting access fails its tests.

## #83 criteria met here

| Criterion | |
|---|---|
| Release and expiry actively close existing streams on every API replica before target cleanup proceeds | ✅ |
| Idle/duration/concurrency limits terminate operations deterministically | ⚠️ duration and concurrency; idle applies to streams, which land with the transport |
| Cached contexts cannot outlive lease expiry, revocation, or a UID change | ✅ registration is UID-keyed and re-resolved per request |

## Not in this PR — the transport

TTY attach and port-forward themselves. They need the Kubernetes upgrade protocols plumbed through axum, terminal resize, signal forwarding, and idle/byte limits per stream. That is the larger half of #83 and it lands next; **this** half is what makes it safe to add, because a stream you cannot revoke is one you should not open.

Also still open from #81, and unchanged here: child placement for logs and exec is refused explicitly rather than read from the wrong cluster.

### DIFF
--- a/src/api/mod.rs
+++ b/src/api/mod.rs
@@ -5,4 +5,5 @@ pub mod routes;
 pub mod sandbox;
 pub mod sandbox_access;
 pub mod sandbox_credentials;
+pub mod sandbox_streams;
 pub mod upgrade;
--- a/src/api/sandbox.rs
+++ b/src/api/sandbox.rs
@@ -126,15 +126,50 @@ async fn sandbox_exec<B: ClusterBackend>(
         Err(denied) => return access_denied(&identity, &id, "exec", denied),
     };
 
-    match access::exec_in_sandbox(
-        &scoped,
-        &target,
-        &container,
-        &request.command,
-        SANDBOX_EXEC_TIMEOUT,
-    )
-    .await
-    {
+    // Registered for the duration, so releasing or expiring the lease cancels
+    // the command rather than letting it run on inside a workload that is
+    // being torn down. The guard deregisters on every exit path, including a
+    // panic — a leaked registration would later report cancelling something
+    // that ended long ago.
+    let streams = crate::api::sandbox_streams::registry();
+    if !crate::api::sandbox_streams::may_start_stream(streams, &target.lease_uid).await {
+        return access_denied_with(
+            &identity,
+            &id,
+            "exec",
+            "concurrency_limit",
+            StatusCode::TOO_MANY_REQUESTS,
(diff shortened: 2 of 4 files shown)


jleni:

PR #129, opened 2026-08-12
access(sandbox): add bounded TTY attach and declared-port forwarding

emmanuelm41:
Transport half of #83 · **stacked on #128**

Two things a caller can want that a request/response API cannot give them: a terminal, and a TCP connection to a declared port. Both need a stream — and a stream needs limits, a framing, and a way to be closed by somebody other than the caller. #128 supplied the last of those. This is the transport.

## Not a Kubernetes proxy

The caller's WebSocket **never reaches the API server**. Kobe terminates it, resolves the target through #81, opens a *separate* connection with a credential scoped to one Pod, and copies bytes between the two.

Nothing the caller sends becomes part of a Kubernetes request: not a path, not a header, not a protocol negotiation. That is the difference between "exec into your Sandbox" and "an authenticated tunnel to the API server", and it is why the framing is Kobe's rather than a passthrough.

## Framing

Binary frames, leading channel byte, matching Kubernetes' own `v4` convention so tooling written against one reads naturally against the other:

| | | |
|---|---|---|
| `0` | stdin | caller → sandbox |
| `1` | stdout | sandbox → caller |
| `2` | stderr | sandbox → caller |
| `3` | error | sandbox → caller — a bounded reason, never a raw response |
| `4` | resize | caller → sandbox |

**Everything a caller may send is enumerated; anything else ends the stream.** Silently ignoring a frame leaves a client believing it resized a terminal, or sent input, that never arrived. A caller writing to stdout or the error channel is either confused or probing, and neither is a reason to guess. Text frames are refused for the same reason — two encodings for stdin means a client that guesses wrong sends silently-mangled bytes to a shell.

## Limits, each for its own reason

- **idle** — an abandoned terminal (somebody closed a laptop) holds a connection to the target cluster with nothing to signal that anything is wrong.
- **duration** — the idle bound does not constrain an *active* stream, which would otherwise live as long as the lease and defeat having a transport bound at all.
- **bytes** — `yes > /dev/null` is a one-word way to make the operator relay unbounded traffic, from inside a sandbox whose occupant is by construction not trusted.
- **concurrency** — checked **before** the upgrade, so a caller over the limit gets a status rather than a socket that closes immediately.

The loop sleeps until the *nearest* bound rather than polling, so an idle session costs nothing until it is closed, and the reported reason always comes from re-checking which rule actually fired rather than from which branch woke up.

`StreamLimits` takes its start instant as a parameter: a type that read the clock internally could only be tested against the clock, which is how limit bugs stay hidden until production.

## Other decisions worth review

**Resolution happens before the handshake completes.** A denial has to be an HTTP status the caller's client understands, not a close frame delivered a moment after a successful-looking upgrade.

**Resize on a session with no terminal fails loudly**, as does input to a session with no stdin — #83 asks for exactly this, and the alternative is dropped keystrokes that get blamed on the workload.

**A zero terminal dimension is refused, not forwarded.** Making the workload's own tty handling cope with it is precisely the sort of input a sandboxed process should never receive from the platform itself.

**Asking for a TTY and separate stderr is a contradiction** — a TTY merges them at the kernel level — so `stderr` follows from `tty` rather than being independently selectable.

**End reasons are bounded codes.** A backend message could carry namespaces, node names, or other tenants' object names, so the code carries the whole distinction on its own: a client that cannot tell "you were idle" from "your lease ended" cannot decide whether reconnecting is sensible.

**Port-forward resolves only pool-declared ports.** Alternate spellings of a *declared* port (`03000`) are accepted deliberately — the allowlist is checked against the resolved number, not the string, and refusing them would reject a request for a port the pool published.

## Chart

`pods/attach` and `pods/portforward` join `pods/exec`.

## Testing

```
cargo test --workspace                                  # 1479 passed, 0 failed
```

Mutation-checked: ignoring unknown inbound channels, and removing the byte ceiling, each fail their tests.

## Against #83's acceptance criteria

| Criterion | |
|---|---|
| Exact upgrade protocols without exposing a general API path | ✅ Kobe terminates the socket; nothing passes through |
| Only the owned, Ready, unexpired lease's exact Pod UID, container, declared port | ✅ via #81 |
| Operation timeout, idle timeout, max duration, byte limits, concurrency | ✅ |
| Resize and bounded signals where supported, explicit failure where not | ⚠️ resize yes; **signal forwarding is not implemented** — it needs a channel the Kubernetes exec protocol does not define, so it belongs with a runner contract (#82) |
| Another identity receives 404 and cannot operate or discover a stream | ✅ |
| Callers cannot choose namespace, Pod, container, undeclared port, path, protocol, header | ✅ |
| Release/expiry closes existing streams on every replica | ✅ (#128) |
| Cached contexts cannot outlive expiry, revocation, or a UID change | ✅ |
| Direct and child placements expose identical behaviour | ❌ child is refused explicitly — see below |
| Existing ClusterLease exec/attach/port-forward compatibility unchanged | ✅ untouched |

## Not in this PR

**Child placement.** Interactive access to a composed child cluster is refused with `501` rather than read from the management cluster, which would find nothing or — worse — somebody else's Pod. Closing it means giving the API server a path to the child's kubeconfig, which is #74's composition seen from the request side rather than the controller's. It is the last gap between the two placements and it blocks #76.

**Signal forwarding.** No channel exists for it in the Kubernetes exec protocol; it needs a runner-side contract, which is #82.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -696,6 +696,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "8b52af3cb4058c895d37317bb27508dccc8e5f2d39454016b297bf4a400597b8"
 dependencies = [
  "axum-core",
+ "base64 0.22.1",
  "bytes",
  "form_urlencoded",
  "futures-util",
@@ -714,8 +715,10 @@ dependencies = [
  "serde_json",
  "serde_path_to_error",
  "serde_urlencoded",
+ "sha1",
  "sync_wrapper",
  "tokio",
+ "tokio-tungstenite",
  "tower",
  "tower-layer",
  "tower-service",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -56,7 +56,7 @@ kube = { version = "3.0", features = ["runtime", "derive", "client", "ws"] }
 k8s-openapi = { version = "0.27", features = ["v1_32", "schemars"] }
 
 # Web framework
-axum = { version = "0.8", features = ["json", "matched-path"] }
+axum = { version = "0.8", features = ["json", "matched-path", "ws"] }
 tower = "0.5"
 tower-http = { version = "0.6", features = ["cors", "trace"] }
 
--- a/charts/kobe/templates/rbac.yaml
+++ b/charts/kobe/templates/rbac.yaml
@@ -79,7 +79,7 @@ rules:
   # from the `SandboxPool` spec, which only an administrator can write, and a
   # lease selects a pool and nothing else. No caller input reaches this.
   - apiGroups: [""]
-    resources: ["pods/exec"]
+    resources: ["pods/exec", "pods/attach", "pods/portforward"]
     verbs: ["create"]
   # Sandbox log reads (#81). Separate from `pods` above because `pods/log` is
   # its own resource to RBAC — and worth stating explicitly: this is the verb
--- a/src/api/mod.rs
+++ b/src/api/mod.rs
@@ -6,4 +6,5 @@ pub mod sandbox;
 pub mod sandbox_access;
 pub mod sandbox_credentials;
 pub mod sandbox_streams;
+pub mod sandbox_transport;
 pub mod upgrade;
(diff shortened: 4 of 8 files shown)


jleni:

PR #130, opened 2026-08-12
access(sandbox): serve every operation against composed child clusters

emmanuelm41:
Closes the placement gap in #81/#83 · **stacked on #129**
Epic: #70 — Agent Sandbox · Parent: #75 · Unblocks #76

Every Sandbox operation — logs, exec, attach, port-forward — refused child placement with `501` rather than reading from the management cluster, where the Pod does not exist. That refusal was correct (a lookup there finds nothing, or worse, somebody else's Pod under the same name) but it left the two placements visibly different, which is exactly the equivalence #76 sets out to prove.

Both now resolve through one path. **Child composition changes which cluster the Pod is in; it changes nothing about what the caller may do.**

## A child cluster is never resolved by name

Three fences, and each catches a different real substitution:

- **The internal `ClusterLease` must still exist under the UID recorded at composition.** A later Sandbox's composition reusing the name is not this caller's cluster.
- **Its binding must still name the recorded instance.** A recycled cluster keeps the lease name and replaces the instance underneath it — the substitution the recorded UID exists to catch.
- **Provenance without UIDs is refused before anything is looked up.** Two absent UIDs compare equal, so a check written against them would accept any cluster at all.

## Credentials are minted where the Pod is

A token issued by the management cluster would not authenticate in a child at all, and reaching the child as its *admin* identity is precisely what per-lease credentials replace. So `scoped_client` now takes the resolved cluster rather than the operator's own — the endpoint and trust anchors come from the child's kubeconfig, the identity from a token minted there and scoped to one Pod.

The child kubeconfig is read into memory and never leaves it. Its read error is swallowed rather than propagated, because the error context can carry the Secret's own contents and the value reaches a caller-facing status.

## Not cached, deliberately

A cache would have to be invalidated on lease expiry, revocation, and any UID or provenance change — #83 requires exactly that — and a cache that outlives one of those hands a caller access their lease no longer grants. Per-request resolution costs a `get` and a `Secret` read; that is the right trade until it measurably is not.

## Testing

```
cargo test --workspace                                  # 1482 passed, 0 failed
```

Mutation-checked: resolving the child by name alone — ignoring either the recorded lease UID or the recorded instance UID — fails its tests. The recycled-cluster test also asserts that **no kubeconfig Secret was read**, so a wrong resolution could not have minted a credential against the wrong cluster before failing.

## What this unblocks

| | |
|---|---|
| #81: management and child resolve through the same typed interface | ✅ now true end to end, not just in the resolver |
| #83: direct and child placements expose identical TTY and port-forward behaviour | ✅ |
| #76: dual-placement conformance | now has something to conform |

### DIFF
--- a/src/api/sandbox.rs
+++ b/src/api/sandbox.rs
@@ -120,7 +120,7 @@ async fn prepare_upgrade<B: ClusterBackend>(
         return Err(StatusCode::NOT_FOUND.into_response());
     }
 
-    let (_lease, target) =
+    let (lease, target) =
         match access::resolve_sandbox_target(&state.client, &state.namespace, id, identity).await {
             Ok(resolved) => resolved,
             Err(denied) => return Err(access_denied(identity, id, operation.as_str(), denied)),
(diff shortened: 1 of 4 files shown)


jleni:

PR #131, opened 2026-08-12
execution(sandbox): durable idempotent commands with an honest unknown outcome

emmanuelm41:
Core of #82 · **stacked on #130**

A Kubernetes exec is a *connection*. If it drops, there is no way to answer the only question that matters on a retry: **did my command already run?** An agent that re-runs `terraform apply` because its connection blipped has done something far worse than fail.

## The order is the entire guarantee

```
reserve the idempotency key  →  spawn  →  record the outcome
```

Spawning first and recording afterwards is the obvious implementation, and it is wrong in exactly the case that matters: a crash between the two leaves no trace that anything ran, so the retry runs it again.

Reserving first inverts that. A crash leaves a record saying "this may have run", the retry finds it, and nobody spawns twice. The cost is that some executions end in `Unknown` — which is the honest answer, and the one #82 asks for.

**The reservation is a `create`**, on a name derived from lease UID and key — not a read-then-write. Two concurrent retries race for one object and the API server picks a winner. Read-then-write is the classic way to get two spawns from one idempotency key.

## What is deliberately not stored

Not the argv, not stdin, not the environment, not the output — only a **digest** of the request. Command lines carry secrets routinely (a token in a flag, a connection string in an argument), and Kubernetes object status is readable by anyone with `get` on the type, replicated to every etcd member, and included in backups. The digest still answers "is this the same command as last time?" without ever holding the command.

The digest is **length-prefixed**, so `["a","bc"]` and `["ab","c"]` cannot collide. Without that, a caller could substitute a different command under a key whose first use they already have a result for.

## Conflict, not silent reuse

Same key + same digest → the original execution. Same key + **different** digest → `409`. Running it would give the caller two commands where they asked for one; returning the first one's result would answer a question they did not ask.

A key belonging to **another lease** is simply not found. Reporting a conflict would confirm the other lease exists.

## `Unknown` is a first-class outcome

Declared on a **persisted** deadline — so a restarted controller reaches the same verdict as the one that started the execution — and never guessed. An unreadable deadline waits, because polling is recoverable and a wrong `Unknown` is not.

Not `Failed`, which invites retrying something that may already have had effects. Not `Succeeded`, which would be a lie. `Unknown` is the state that tells a caller the decision is theirs.

**Only `Queued` proves a command did not run.** That is the distinction a caller needs to decide whether retrying is safe, and it is why every other state — including `Unknown` — refuses to spawn again.

## Other decisions

- **Terminal is terminal.** A settled execution that could move again would make every answer Kobe already gave about it provisional, including ones a caller acted on.
- **An exit code is a claim the process ran to completion**: required for `Succeeded`/`Failed`, forbidden for `Cancelled`/`TimedOut`/`Unknown`. A synthesised zero would be indistinguishable from a real success.
- **`cwd` is an absolute path with no nul**, and never implemented with a shell — `cd X && ...` would make quoting the security boundary. The runner validates it too; this is the sanity check, not the enforcement.
- **Execution names are hashed** from lease UID and key, so a caller-supplied key cannot escape into an object name, exceed the length limit, or collide across leases.
- **Records are owner-referenced to their lease**, so history ends with the Sandbox rather than describing a workload nobody can inspect.

An **execution reaper** settles records left `Running` by a process that disappeared. It runs on every replica and is idempotent — a settled execution is never re-judged — so several replicas reaching the same verdict is a no-op.

## Testing

```
cargo test --workspace                                  # 1509 passed, 0 failed
```

Mutation-checked: ignoring a digest conflict, collapsing the digest's argument boundaries, letting a running execution spawn again, and declaring `Unknown` without a deadline — each fails its tests.

## Two deviations, stated plainly

**The key is reserved in Kobe's API, not in the target.** #82 asks for the target. Doing so needs a writable, durable path inside a container Kobe does not control — the runner contract this issue itself defers — and the difference would only matter if two independent Kobe deployments could drive one Sandbox, which they cannot. Reserving in Kobe is durable, atomic, and available today.

**Detached execution is refused with `501`.** It needs a supervisor inside the container that outlives the connection. A "detached" execution that actually dies with its connection is *worse* than none, because a caller builds on the guarantee it appears to offer. Wait mode is fully implemented and carries every idempotency guarantee above.

## Against #82's acceptance criteria

| Criterion | |
|---|---|
| Wait mode returns distinct stdout/stderr and the exit status | ✅ |
| Reusing one key cannot start a duplicate across retries or restarts | ✅ |
| Conflicting content for one key is rejected | ✅ |
| Lost/unverifiable outcomes report `Unknown`, never success or auto-retry | ✅ |
| `cwd` validated, never an implicit shell | ✅ |
| Output, arguments, stdin, environment, credentials absent from status/events/logs | ✅ |
| Foreign identity, stale UID, release, expiry, quarantine deny | ✅ via #81 |
| Resource/output/concurrency bounds | ✅ output, timeout, concurrency |
| Detached survives disconnect; poll, log tail, cancel | ⚠️ poll and cancel yes; **detach and log tail need the runner** |
| Failure injection across crash-before-spawn / after-spawn / after-ack | ⚠️ the *states* are covered by unit tests; injected crashes need the integration harness in #76 |
| Timeout/cancel terminates the process group | ⚠️ cancel is recorded and the connection dropped; **killing the process group needs the runner** |

### DIFF
--- a/charts/kobe/crds/sandboxexecutions.yaml
+++ b/charts/kobe/crds/sandboxexecutions.yaml
@@ -1,0 +1,135 @@
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  name: sandboxexecutions.kobe.kunobi.ninja
+spec:
+  group: kobe.kunobi.ninja
+  names:
+    categories: []
+    kind: SandboxExecution
+    plural: sandboxexecutions
+    shortNames:
+    - sbxe
+    singular: sandboxexecution
+  scope: Namespaced
+  versions:
+  - additionalPrinterColumns:
+    - jsonPath: .status.state
+      name: State
+      type: string
+    - jsonPath: .status.exitCode
+      name: Exit
+      type: integer
+    - jsonPath: .metadata.creationTimestamp
+      name: Age
+      type: date
+    name: v1alpha1
+    schema:
+      openAPIV3Schema:
+        description: Auto-generated derived type for SandboxExecutionSpec via `CustomResource`
+        properties:
+          spec:
+            description: |-
+              One command's durable record.
+
+              Owner-referenced to its `SandboxLease`, so it cannot outlive the lease it
+              belongs to: #82 requires history to end with the Sandbox, and an execution
+              record that survived would describe a workload nobody can inspect.
+            properties:
+              detached:
+                default: false
+                description: Whether the caller waited for the result.
+                type: boolean
+              idempotencyKey:
+                description: Caller-supplied key. Two requests with this key are the same request.
+                maxLength: 253
+                minLength: 1
+                type: string
+              leaseUid:
+                description: |-
+                  The lease this ran under, by UID. An execution never outlives its exact
+                  lease: a same-named successor is a different Sandbox, and answering its
+                  caller with this record would describe somebody else's command.
+                minLength: 1
+                type: string
+              podUid:
+                description: |-
+                  The Pod it ran in, by UID. Recorded so a replaced Pod invalidates the
+                  record rather than silently re-pointing it.
+                minLength: 1
+                type: string
+              requestDigest:
+                description: |-
+                  SHA-256 over the canonical request. Never the request itself: a command
+                  line routinely carries secrets, and object status is readable by anyone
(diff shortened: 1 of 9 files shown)


jleni:

PR #132, opened 2026-08-12
cli(sandbox): add run, exec, logs and cancel

emmanuelm41:
Core of #84 · **stacked on #131**
Epic: #70 — Agent Sandbox · Parent: #75 · Unblocks #76

```bash
kobe sandbox run agents --ttl 30m -- /agent analyse ./repo
kobe sandbox exec sbx-7f3a --cwd /work -- ./build.sh
kobe sandbox logs sbx-7f3a --tail 500
kobe sandbox cancel sbx-7f3a --execution sbxe-91c4
```

## Two audiences, one contract

A human runs `exec` and reads the output. An agent runs the same command with `--output json` and parses it. **The agent is the harder consumer and it decides the design**: exact remote exit code, stdout and stderr kept apart, and a transport failure that cannot be mistaken for a command failure.

## Exit codes carry meaning

The remote command's exit code **is** this process's exit code. That is the whole point of `exec` inside a script — `set -e` has to work — so Kobe's own failures must not be able to collide with it.

They exit `125`: the convention `docker run` and `env` already use for "the tool failed, not your command", so a script that already handles it needs no new knowledge. It is reported when an execution ends *without* an exit code — `Unknown`, `Cancelled`, `TimedOut`. Reporting `0` there would tell a caller their command succeeded when nobody knows whether it started.

## `run` cleans up, and says so separately

`run` composes create → wait → execute → release, and the release is attempted on **every** terminal path. A lease created and abandoned is the worst thing this command can do: nothing else notices until the TTL reaps it, and the caller is billed for the gap.

Cleanup failure is reported **beside** the command's result, never folded into it — collapsing them would either hide a leaked lease behind a successful command, or fail a command that actually worked:

```json
{
  "apiVersion": "kobe.sandbox/v1",
  "state": "Succeeded",
  "exitCode": 0,
  "cleanup": { "released": false, "error": "HTTP 503" }
}
```

A run whose command succeeded but whose lease leaked exits `125` rather than `0`, so an unattended caller notices something it would otherwise have to go looking for.

## No shell

`argv` is sent as-is. There are no quoting rules of Kobe's own to learn, and no way for an argument to become a shell operator. A caller who wants a shell asks for one — and its quoting is then theirs to reason about:

```bash
kobe sandbox exec sbx-7f3a -- /bin/sh -lc 'a && b'
```

## Retries are safe

Every invocation carries a fresh idempotency key, generated **per command** rather than derived from the argv. A derived key would make two deliberate runs of the same command collide, and the second would silently return the first one's result — a wrong answer, not a cached one. Generated without adding a dependency, since the key scopes a retry rather than authorising anything.

## Machine output

Versioned by `apiVersion`, which changes only on a breaking change — a field that changed meaning without one would break an agent silently, at some later date, with no way to tell which side was wrong. `stdout` and `stderr` are always distinct fields.

`cancel` reports the state that **actually** applies: an execution that had already finished is reported as finished, because saying "cancelled" would be a claim the caller might act on.

## Documentation

`docs/kobe-docs/commands/sandbox.mdx`, with a headless-agent section: the exit-code table, a `jq` example that checks cleanup, and why `Unknown` is deliberately not reported as a failure.

## Testing

```
cargo test --workspace                                  # 1514 passed, 0 failed
```

Mutation-checked: reporting `0` for an execution that never completed, and folding cleanup into the command result, each fail their tests.

## Against #84's acceptance criteria

| Criterion | |
|---|---|
| Commands work identically for management and child placements | ✅ the server resolves placement (#130); the CLI never sees it |
| Machine-readable output is deterministic and documented | ✅ |
| `exec`/`run` preserve the exact remote exit code while separately reporting transport/release failures | ✅ |
| `run` attempts release on every terminal path | ✅ |
| Retry uses idempotency keys and cannot duplicate a command | ✅ |
| Foreign, missing, stale, expired, released, quarantined targets fail closed | ✅ server-side (#126) |
| No command downloads kubeconfig or exposes target credentials | ✅ |
| Public documentation includes examples for headless agents | ✅ |
| `attach` / `port-forward` subcommands | ❌ **not here** — see below |
| Alias resolution scoped to the caller, rejecting ambiguity | ❌ **not here** |

## Not in this PR

**`attach` and `port-forward` subcommands.** The server side landed in #129, but the client half is a terminal implementation — raw mode, resize propagation, signal handling, a local listener for forwarding — and it is a materially different piece of work from these four. Shipping a half-working TTY is worse than not shipping one: a terminal that mangles input is discovered mid-session, on somebody's real work.

**Alias resolution.** `SandboxLease.spec.alias` exists and admission reserves it, but nothing resolves an alias to a lease id yet. That is a server-side lookup with an ambiguity rule, not a CLI concern, and it belongs with the endpoint that serves it.

Both are called out rather than glossed; neither weakens what is here.

### DIFF
--- a/crates/kobectl/src/commands/mod.rs
+++ b/crates/kobectl/src/commands/mod.rs
@@ -9,6 +9,7 @@ mod picker;
 mod pools;
 mod purge;
 mod release;
+pub(crate) mod sandbox;
 mod select;
 pub(crate) mod session;
 mod state;
--- a/crates/kobectl/src/commands/sandbox.rs
+++ b/crates/kobectl/src/commands/sandbox.rs
@@ -1,0 +1,743 @@
+//! `kobe sandbox` — exec, logs, cancel, and run (#84).
+//!
+//! # Two audiences, one contract
+//!
+//! A human runs `kobe sandbox exec` and reads the output. An agent runs the
+//! same command with `--output json` and parses it. The agent is the harder
+//! consumer and the one that decides the design: it needs the **exact remote
+//! exit code**, stdout and stderr kept apart, and a transport failure that
+//! cannot be mistaken for a command failure.
+//!
+//! # Exit codes carry meaning
+//!
+//! The remote command's exit code is this process's exit code. That is the
+//! whole point of `exec` in a script — `set -e` has to work — so Kobe's own
+//! failures must not be able to collide with it. They exit `125`, chosen
+//! because it is what `docker run` and `env` use for exactly this: "the tool
+//! failed, not your command".
+//!
+//! # `run` cleans up, and says so separately
+//!
+//! `run` creates a lease, executes, and releases. The release is attempted on
+//! every terminal path — success, non-zero exit, timeout, interruption — and
+//! its failure is reported *separately* from the command's result. Collapsing
+//! them would either hide a leaked lease behind a successful command, or fail
+//! a command that actually worked.
+
+use std::io::Write;
+
+use anyhow::{Context, Result};
+use serde::{Deserialize, Serialize};
+
+use super::config::{CliConfig, ResolvedConfig};
+use super::{OutputFormat, authed_client, get_auth_header, print_json, with_auth};
+
+/// Exit code for Kobe's own failures.
+///
+/// A remote command's exit code is passed through verbatim, so Kobe needs a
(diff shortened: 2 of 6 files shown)


jleni:

PR #133, opened 2026-08-12
access(sandbox): resolve caller-scoped aliases, refusing ambiguity

emmanuelm41:
Completes #84's alias criterion · **stacked on #132**

```bash
kobe sandbox exec dev -- /agent status     # `dev` is your alias, not a lease id
```

`SandboxLease.spec.alias` existed and admission reserved it. Nothing turned one into a lease id. Now every command that takes a lease takes an alias.

## Scoped to the caller

Aliases are **caller-chosen**, so two tenants picking `dev` is the expected case rather than a collision to break arbitrarily. Resolution is scoped so one tenant's alias can never name another's lease.

**The label selector is a prefilter, not the authorisation.** The requester hash narrows the list; each candidate is then re-checked against the complete principal tuple. The hash is computed over caller-influenced values, so a collision would otherwise be enough to reach another tenant's lease by aliasing it. The answer is `NotFound` rather than a conflict — a conflict would confirm the other lease exists.

## Ambiguity is refused, not guessed

Two live leases under one alias → `409`. Picking the newest would land the caller's next command somewhere they did not expect, and they would find out from its **side effects**. `409` rather than `400` because the request is well-formed: the fix is to their leases, not their command.

## Terminal leases do not shadow live ones

Reusing an alias after releasing is the ordinary workflow. If released, expired or quarantined leases counted, every *second* use of an alias would be "ambiguous" — the feature would fail exactly when it is most wanted.

## Ids and aliases are told apart by shape

Anything starting with `sbx-` is a lease id; anything else is an alias. Never "try one, fall back to the other": a fallback would let a caller who names a lease that has just expired silently reach a **different** lease that happens to use that name as an alias.

## Testing

```
cargo test --workspace                                  # 1519 passed, 0 failed
```

Mutation-checked: guessing an ambiguous alias, trusting the label prefilter as authorisation, and counting terminal leases — each fails its tests.

Documented in `docs/kobe-docs/commands/sandbox.mdx`.

### DIFF
--- a/docs/kobe-docs/commands/sandbox.mdx
+++ b/docs/kobe-docs/commands/sandbox.mdx
@@ -135,6 +135,31 @@ may have run. It is deliberately **not** reported as a failure, because a
 failure invites an automatic retry of something that may already have had
 effects. Deciding what to do is yours.
 
+## Aliases
+
+Any command that takes a lease id also takes an **alias** — the name you gave
+the sandbox when you created it:
+
+```bash
+kobe sandbox exec dev -- /agent status
+```
+
+Aliases are yours. Two people using `dev` never collide: resolution is scoped
+to the caller, so an alias can only ever name one of your own sandboxes.
+
+**Ambiguity is refused, not guessed.** If two of your live sandboxes share an
+alias, the command fails with `409` rather than picking one — a silent choice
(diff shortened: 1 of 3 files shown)


jleni:

PR #134, opened 2026-08-12
cli(sandbox): add attach and port-forward

emmanuelm41:
Completes #84 · **stacked on #133**

```bash
kobe sandbox attach dev                    # interactive session
kobe sandbox attach dev -- /bin/bash       # start a shell instead
kobe sandbox port-forward dev 8080:http    # forward a declared port
```

The server side landed in #129. This is the terminal half.

## ⚠️ A latent panic this PR had to fix first

The WebSocket client's TLS feature compiles a **second** rustls crypto provider (aws-lc-rs) alongside reqwest's ring. With two present, rustls refuses to choose and **panics on the first TLS connection** — which would have broken every existing `kobe` command, not just the new ones.

`main` now installs ring explicitly, exactly as the operator already does. Worth reviewing on its own: it is a one-line change guarding a whole-CLI regression.

## Raw mode is a promise to restore it

`attach` puts the terminal in raw mode so **Ctrl-C reaches the workload** rather than killing the CLI — that is the whole reason for raw mode. It restores on every exit path **including a panic**: a process that dies in raw mode leaves a terminal that does not echo, and the user's next instinct is to close the window rather than type `reset` blind. The panic hook restores *before* the message prints, because a panic rendered in raw mode arrives as a staircase.

**The initial window size is sent before anything is typed.** A shell that starts believing the terminal is 80×24 renders its first prompt wrong, and no later resize event arrives to correct it if the window never changes. Subsequent resizes are forwarded.

**Unmapped keys send nothing rather than something wrong.** Doing nothing is a visible non-event; sending the wrong byte is one the workload acts on.

## Asymmetric tolerance, on purpose

The client **ignores** unknown server channels; the server **refuses** unknown client frames. That asymmetry is deliberate: a server that adds a channel must not break clients that never needed it — but a server ignoring a client's frame would leave that client believing it sent a resize, or a keystroke, that never arrived.

## A cut-off session must not look like a clean one

`completed` and `closed` exit `0`. A revoked lease, a limit, or a reason from a newer server exits `125` with the reason printed:

```
kobe: session ended: idle_timeout
```

An unattended caller has to tell "ended normally" from "cut off", because only one of those means the work finished.

## The URL scheme is switched, not substituted

`replace("http", "ws")` rewrites the first occurrence *anywhere* in the URL — including a host called `http-gateway.internal` — and would silently point the caller somewhere else. The scheme is set through the parsed URL instead.

## port-forward

- **Only pool-declared ports resolve.** Without that a forward is a general tunnel into the sandbox's network namespace, reaching a debug listener or a metrics endpoint the administrator never meant to publish.
- **Binds loopback by default.** A forward reachable from the network turns "a port on my machine" into "a port on the office LAN", and the sandbox behind it belongs to one caller.
- **One connection at a time.** Each local connection needs its own upstream stream, and those count against the lease's concurrency limit — a browser opening six sockets would exhaust it, and the failures would look like the sandbox misbehaving.
- The remote half of `LOCAL:REMOTE` stays a **string**, so `8080:http` works: parsing it as a number would refuse the form an administrator publishing named ports intends people to use.

## Testing

```
cargo test --workspace                                  # 1527 passed, 0 failed
```

Mutation-checked: substituting the URL scheme, treating every end reason as clean, and dropping the control-key translation — each fails its tests.

## #84 is now complete

| Criterion | |
|---|---|
| TTY resize/signals work through the restricted API only | ✅ |
| Declared-port forwarding | ✅ |
| Another identity cannot operate or discover a stream | ✅ server-side |
| Callers cannot choose namespace, Pod, container, undeclared port | ✅ server-side |
| Alias resolution scoped to the caller, rejecting ambiguity | ✅ (#133) |
| Public documentation with headless-agent examples | ✅ |

> What is **not** covered by unit tests is the live WebSocket round-trip and real terminal behaviour — the framing, key encoding, URL derivation and exit-code rules are, but an actual attached session needs a cluster. That belongs to #76's conformance suite.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1107,6 +1107,7 @@ checksum = "829d955a0bb380ef178a640b91779e3987da38c9aea133b20614cfed8cdea9c6"
 dependencies = [
  "bitflags",
  "crossterm_winapi",
+ "futures-core",
  "mio",
  "parking_lot",
  "rustix 0.38.44",
@@ -2729,14 +2730,17 @@ dependencies = [
  "clap",
  "crossterm",
  "dirs",
+ "futures-util",
  "kunobi-auth",
  "ratatui",
  "reqwest 0.12.28",
+ "rustls 0.23.37",
  "serde",
  "serde_json",
  "serde_yaml_ng",
  "tempfile",
  "tokio",
+ "tokio-tungstenite",
  "toml",
  "url",
 ]
@@ -5372,8 +5376,12 @@ checksum = "d25a406cddcc431a75d3d9afc6a7c0f7428d4891dd973e4d54c56b46127bf857"
 dependencies = [
  "futures-util",
  "log",
+ "rustls 0.23.37",
+ "rustls-pki-types",
  "tokio",
+ "tokio-rustls 0.26.4",
  "tungstenite",
+ "webpki-roots 0.26.11",
 ]
 
 [[package]]
@@ -5629,6 +5637,8 @@ dependencies = [
  "httparse",
  "log",
  "rand 0.9.2",
+ "rustls 0.23.37",
+ "rustls-pki-types",
  "sha1",
  "thiserror 2.0.18",
  "utf-8",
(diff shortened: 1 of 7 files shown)


jleni:

PR #135, opened 2026-08-12
conformance(sandbox): prove dual-placement isolation, expiry, and cleanup

### LINKED ISSUE #76: conformance(sandbox): prove dual-placement isolation, expiry, and cleanup

emmanuelm41:
Closes #76 · **stacked on #134** · the closure gate for #70
Epic: #70 — Agent Sandbox

## One suite, two placements

Every scenario runs **twice**: management placement and child-cluster placement. That is the whole design. Two copies of a suite would prove nothing about equivalence — they would drift, and the drift would be invisible until somebody's child-placed Sandbox behaved differently from the one they tested against.

So each scenario takes a `Placement` and *nothing else changes*: same requests, same assertions, same expected status shapes.

**A test guards exactly that.** A scenario declared with `#[tokio::test]` instead of through the `both_placements!` macro fails the suite, because it would run against one placement only. Mutation-checked by writing such a scenario and confirming the guard fires — that is the one mistake that would quietly gut this file.

## Only the public contract

Everything goes through the HTTP API a caller actually uses. Nothing reads a CRD, a Secret, or the child cluster's kubeconfig.

Partly because those are not the contract, and partly because a suite that reached around the API could pass while the API itself was broken. But mostly because it lets the suite assert something a Kubernetes-level test **cannot**: that a caller is *unable* to reach the cluster underneath. A test holding a kubeconfig cannot prove the absence of one.

## Eleven scenarios, twenty-two runs

| | |
|---|---|
| Ready with the same typed shape | …and neither response mentions the composition underneath it |
| **TTL starts at readiness** | not at request — for child placement, where provisioning means *building a cluster*, the difference could be most of the lease |
| Exact remote exit code | stdout/stderr distinct; a non-zero exit is the caller's result, not an error |
| **One key cannot run twice** | asserted through a command with a *side effect*, so a duplicate spawn is observable rather than suspected; a reused key with different content conflicts |
| Cross-tenant denial | every operation 404s — never 403, which would confirm the lease exists and allow enumeration by guessing ids |
| Release rejects further access | including from a caller who was mid-session |
| No target credentials | no certificate, kubeconfig, token, or internal cluster identity on any caller-readable surface |
| **Workload cannot escape** | checked from *inside* the sandbox, because that is where an attacker would be |
| Undeclared ports refused | |
| Tampered target fails closed | caller-chosen container, smuggled request fields |
| Cancel while provisioning | settles without quarantining — for child placement, that a partial cluster allocation is not stranded |

## Two decisions worth review

**The cross-tenant scenarios fail without a second identity — they never skip.** A conformance suite that silently drops its isolation checks reports success while proving nothing about the property those checks exist for. That is worse than a suite that does not run.

**Each scenario releases its sandbox even on panic.** A leaked sandbox makes the *next* scenario queue and fail for an unrelated reason, which is how one broken assertion becomes a suite nobody can read.

## CI

Wired into the e2e job on the **k3s leg only**: child placement requires a receipt-capable k3s pool, so the other backends have nothing to compare against and would report a skip that reads as a pass.

It skips **loudly**, with a warning annotation, when the harness has published no Sandbox endpoint — rather than failing an e2e run for an environment that predates the Sandbox pools.

```
just test-sandbox-conformance
```

## Testing

```
cargo test --workspace                                  # 1529 passed, 0 failed
```

The suite's own shape guards run in the normal suite; the scenarios are `#[ignore]`d and need a live endpoint.

## What this suite does not yet cover

Stated rather than implied, because #76's matrix is explicit and a checklist that looks complete is worse than one that admits gaps:

- **Controller/API restart at each phase.** The scenarios drive the public API; restarting the operator mid-phase needs harness-level control the e2e script does not expose yet. The *states* those restarts produce (`Unknown`, re-reserved keys, idempotent transitions) are unit-tested in #131.
- **Quarantine on uncertain teardown**, and **target loss with cleanup retry.** Both require injecting a failure into the target — revoking RBAC mid-teardown, or making the child API unreachable. Also harness-level.
- **TTY and port-forward round-trips.** The framing, key encoding, URL derivation and exit-code rules are unit-tested in #129/#134; an actual attached session needs a pty in CI.

Each is a harness capability rather than a suite gap, and each is worth its own issue against the e2e script rather than a scenario that cannot run.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -289,6 +289,25 @@ jobs:
           path: /tmp/kind-logs
           if-no-files-found: ignore
 
+      # Dual-placement Sandbox conformance (#76). Runs only on the k3s leg:
+      # child placement REQUIRES a receipt-capable k3s pool, so the other
+      # backends have nothing to compare against and would report a skip that
+      # reads as a pass.
+      #
+      # Skipped LOUDLY when the harness has not published a Sandbox endpoint,
+      # rather than failing the whole e2e run for an environment that predates
+      # the Sandbox pools.
+      - name: Sandbox conformance (management + child)
+        if: matrix.backend == 'k3s'
+        env:
+          KOBE_SANDBOX_E2E: "1"
+        run: |
+          if [ -z "${KOBE_ENDPOINT:-}" ] || [ -z "${KOBE_SANDBOX_POOL_CHILD:-}" ]; then
+            echo "::warning title=Sandbox conformance skipped::no Sandbox endpoint or child pool configured for this environment (#76)"
+            exit 0
+          fi
+          mise exec -- just test-sandbox-conformance
+
       # Always tear down so kind clusters never leak on the persistent runner.
       - name: Tear down the e2e harness
         if: always()
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -148,5 +148,13 @@ regex = "1"
 
 [dev-dependencies]
 tokio-test = "0.4"
+# The conformance suite drives the public HTTP API, including a blocking
+# release from a `Drop` — a leaked sandbox makes the NEXT scenario queue and
+# fail for an unrelated reason.
+reqwest = { version = "0.12", features = [
+  "json",
+  "rustls-tls",
+  "blocking",
+], default-features = false }
 wiremock = "0.6"
 tempfile = "3"
--- a/justfile
+++ b/justfile
@@ -103,6 +103,22 @@ test-smoke-pool pool='ci-small' ttl='2m' *args:
 test-smoke-k3s cluster='e2e-kobe' *args:
     @env E2E_CLUSTER={{ cluster }} mise exec -- bun run ./hack/test-e2e-k3s.ts {{ args }}
 
+# Dual-placement Sandbox conformance (#76).
+#
+# One suite, run against BOTH placements — management and child cluster. That
+# is the point: two copies would drift, and the drift would be invisible until
+# somebody's child-placed Sandbox behaved differently from the one they tested.
+#
+# Serial on purpose: these lease real capacity, and a pool sized for a handful
+# of sandboxes queues rather than fails, which reads as a timeout.
+#
+# KOBE_TOKEN_OTHER is REQUIRED. The cross-tenant scenarios are not optional,
+# and a suite that skipped them would report success while proving nothing
+# about the property they exist for.
+[group('test')]
+test-sandbox-conformance:
+    @KOBE_SANDBOX_E2E=1 cargo test --test sandbox_conformance -- --ignored --test-threads=1
+
 # Local e2e environment entrypoint
 [group('dev')]
 e2e *args:
--- a/tests/sandbox_conformance.rs
+++ b/tests/sandbox_conformance.rs
@@ -1,0 +1,786 @@
+//! Dual-placement Sandbox conformance (#76) — the closure gate for #70.
+//!
+//! # One suite, two placements
+//!
+//! Every scenario here runs **twice**: once against a management-placement
+//! pool, once against a child-cluster pool. That is the entire design. Two
+//! copies of a suite would prove nothing about equivalence — they would drift,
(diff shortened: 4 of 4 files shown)


jleni:

Issue #138, opened 2026-08-13
test(e2e): expose restart, failure-injection and pty control for Sandbox conformance


emmanuelm41:
Split out of #76. The conformance suite (#135) covers eleven dual-placement scenarios, but three rows of #76's matrix need capabilities `hack/e2e.ts` does not expose. Each is a **harness** gap rather than a suite gap — writing them as scenarios today would produce tests that cannot run, which is worse than not having them, because a skipped scenario reads as a passing one.

## 1. Restart the operator at an arbitrary phase

#76 asks for idempotency "after each provenance write, child bind, bootstrap, Claim create, Ready transition, access grant, and teardown step".

The suite drives the public HTTP API and has no way to restart the operator mid-phase, nor to know it has reached a given phase before doing so. It needs something like:

```
e2e.ts restart-operator --after-phase=<provenance|bind|claim|ready|teardown>
```

The **states** those restarts produce — `Unknown`, a re-reserved idempotency key, an idempotent phase transition — are unit-tested in #131 and #125. What is untested is that a real restart at a real moment produces them.

## 2. Inject a failure into the target

#76 asks that "delete/finalizer/PDB/webhook/API-unreachable/storage failures remain visible, retryable, and quarantined", and that quarantine happens on uncertain teardown.

Every one of those needs the target made to fail deliberately: revoking the operator's RBAC mid-teardown, making a child cluster's API unreachable, or wedging a finalizer. The suite cannot do this through the API it tests, and should not — a suite that could break its own target could also mask a break it did not intend.

## 3. A pty for TTY round-trips

`kobe sandbox attach` puts the terminal in raw mode and forwards keystrokes and resizes. The framing, key encoding, URL derivation and exit-code rules are unit-tested (#129, #134), but an actual attached session needs a pty allocated in CI to prove that a keystroke typed on one end arrives at the workload.

## Why these are worth doing

Each of the three covers a failure that only appears under real conditions, and each maps to a property the design leans on:

- restart → "a crash never causes a duplicate spawn";
- injected failure → "uncertain teardown withholds capacity rather than releasing it";
- pty → "Ctrl-C reaches the workload".

All three are asserted by construction today. None is proven end to end.

jleni:
Implemented by PR #143 and delivered to main through PR #160. The E2E harness now exposes operator restart, failure injection, and PTY control. Remaining real child-placement proof stays tracked in #76.

Issue #123, opened 2026-08-12
chore(sandbox): retire terminal SandboxLease objects


emmanuelm41:
Split out of #73, which is otherwise complete.

`Released` and `Expired` leases no longer consume capacity — `consumes_capacity()` is false for both, and #120 releases their quota reservations against verified teardown. So this is **not** a leak. The objects themselves simply accumulate: one per Sandbox ever leased, forever.

Left out of #120 deliberately, because the question is a policy one rather than a code one:

- **How long is a terminal lease worth keeping?** It is the audit record of who held what, for how long, and whether cleanup was proven. Deleting it at completion destroys that; keeping it forever grows `etcd` without bound.
- **Does `Quarantined` retire at all?** It is deliberately terminal and deliberately still consuming capacity. A retention sweep that treated it like the clean phases would silently release capacity that was withheld precisely because nobody could prove it was safe — which is the failure the quarantine exists to prevent. It should almost certainly be exempt, and that should be explicit rather than incidental.
- **Where does the sweep live?** Most likely alongside `run_sandbox_admission_reaper`, which already runs on a durable interval and already reasons about abandoned admission state.

Suggested shape, pending a call on the above: a configurable retention window (default on the order of days), applied to `Released` and `Expired` only, with `Quarantined` explicitly excluded and the exclusion covered by a test.

jleni:
Implemented by PR #141 and delivered to main through PR #160. Clean terminal SandboxLease retention is bounded, while Quarantined leases remain explicitly exempt.

PR #139, opened 2026-08-13
epic(sandbox): Agent Sandbox — full feature integration

emmanuelm41:
**Draft, deliberately.** Everything for #70 accumulates here and nothing merges to `main` until the whole feature is done and proven. The draft exists so CI runs on the complete surface rather than on fragments that individually pass.

Integrates the sixteen PRs #117 → #135, which remain open as the **review units** — each one is a single decision with its reasoning in the body. This branch is where they run together.

## What the merge surfaced

A clean textual merge hid three real signature drifts against what landed on `main` in #115/#116. All three failed to compile, which is the good outcome:

- `release_reservations_for_lease` gained the lease itself;
- `verified_destroy_eligibility` gained the backend, and now refuses non-k3s **itself** — so the duplicate backend check in child composition was removed rather than kept in sync, since two copies of a rule drift and the one that silently wins is whichever runs first;
- `TeardownCheck` gained `verified`, the concrete resources a check actually looked at.

It also surfaced a genuinely wrong **test fixture**: #116 hardened reservation release so a reservation is deletable only if its *name* derives from the lease's own principal — carrying the right UID label is deliberately not enough, or anyone able to create a coordination Lease could label one with a victim's UID and have Kobe's credentials delete it during that victim's cleanup. Four of my tests used a hand-written reservation name and correctly began failing. The fixture now derives the name, so it stays honest if the rule changes again.

## Regression surface

Two of the four foundation PRs already on `main` are **not** purely additive — #96 adds a `Quarantined` phase and #115 adds a finalizer gate, both in the ClusterInstance lifecycle every existing pool runs through. That is the area #119 was about.

They are inert on `main` today: the gate returns early unless the bound lease sets `cleanup_mode: VerifiedDestroy`, which defaults to `Standard`, and nothing on `main` constructs such a lease.

**This branch is where that changes.** Child composition is the only thing that requests `VerifiedDestroy`. From here, the gate is live — and a lease requesting it whose instance has no creation-time teardown plan quarantines rather than releasing. That is deliberate (verifying against a reconstructed plan proves only that the guess was satisfied) but it is a real behaviour change, and it is the thing to look hardest at before this leaves draft.

## State

```
cargo test --workspace                                  # 1571 passed, 0 failed
```

## Still to land here

- terminal `SandboxLease` retention (#123)
- admission rate limiting (#101 remainder)
- e2e harness capabilities the conformance suite needs (#138) — restart-at-phase, failure injection, a pty

## Deliberately out of scope, pending your call

- `managed` mode (#72) and child-cluster runtime bootstrap (#74): both install and retain cluster-scoped CRDs, RBAC and a webhook with cluster-admin. Paused pending approval; child placement validates a runtime the operator installs instead.
- Detached execution (#82): needs a supervisor inside the container that outlives the connection.


jleni:

PR #140, opened 2026-08-13
hardening(sandbox): rate-limit admission per principal, per attempt

emmanuelm41:
Closes the last two open boxes of #101. One is a change; the other is a deliberate non-change, argued and pinned by a test.

Stacked on #135 (`feat/sandbox-conformance`).

## The rate limit

`max_concurrent_leases` bounds what a principal may **hold**. It says nothing about how fast they may **ask**, and the two are not the same bound — because refusing an attempt is not free. By the time admission can answer "no", the handler has already spent a `SandboxLease` CREATE, up to `max_concurrent_leases` reservation CREATEs, and a fenced DELETE to undo them. A caller sitting at their limit can loop on 429s forever, spend none of their own quota, and still saturate the API server every other principal shares. That is the gap the issue names.

So the budget is charged per **attempt**, as the first statement of the handler, and never refunded. A caller whose requests all fail is charged exactly like one whose requests all succeed — which is the whole point, and why the charge is not a decision taken at one of the handler's exits: every exit below has already paid, and none is reachable without passing through it.

A *throttled* attempt is the one thing that costs nothing: it is refused before any I/O, so there is nothing to bill. Charging it too would turn a bounded delay into a permanent lockout for anything with a tight retry loop — punishing bad retry code rather than bounding load.

Token bucket: ten attempts back-to-back, then one every two seconds. Sized so an agent runner opening a handful of sandboxes at once never sees a 429, while a retry loop drops from "as fast as the network allows" to a trickle. The key is `principal_hash` — the same digest that names the quota and alias reservations, so the limiter and the ledger cannot disagree about who a caller is.

Refusals answer with `Retry-After`, rounded up. Without it a throttled client has no signal but "no", and the only strategy left is to poll — which is the load the throttle was raised against.

Two claims stated narrowly rather than implied:

- **Per-process.** With N replicas the effective ceiling is N times these numbers. A cluster-wide limiter would need a write per *rejected* attempt, spending the exact resource it protects; bounding the blast radius by a constant factor is the honest claim.
- **Bounded table.** At the cap it refuses rather than evicting live buckets — forgetting an active principal to make room *is* the evasion, since a flood of distinct credentials would evict the very buckets holding it back.

`kobe_sandbox_admission_rate_limited_total` counts refusals, registered at startup so an alert reads zero rather than "no data".

## `requester_type` stays out of the principal digest

The issue asks to confirm this is intended. It is — and including it would be a quota **bypass**, not a tightening.

`requester_type` is `"{provider}:{matched rule value}"`. It names the AccessPolicy rule that matched *this* request, not the caller. It moves when an admin reorders or edits rules, when the IdP changes a claim, and — decisively — when the same subject presents a token whose claims select a different rule. Feeding it into the digest would make each of those a rename of `sbx-quota-{hash}-{slot}` and `sbx-alias-{hash}-{alias}`.

A rename is not a partition. The old reservations still exist and still consume backend resources, but the new selector cannot see them — so the caller gets their full `max_concurrent_leases` again and can re-take an alias someone is still holding. `principal_hash` already documents that hazard for a *code* change, where it is a one-off migration; including `requester_type` makes it reachable at runtime, repeatedly, by a caller who can influence their own claims.

It would also split the digest from `principal_matches`, which decides ownership on provider/issuer/identity. The label selector would then narrow a caller out of leases they still own, and they could no longer release the very lease holding their slot — the lockout class this epic has spent several PRs removing.

Unchanged, with the reasoning moved into a test so it is a constraint rather than a comment.

## Mutation checks

Each mutation was applied, the named test confirmed to fail, and the tree restored:

| Mutation | Test that caught it |
|---|---|
| charge on success instead of at entry | `a_caller_whose_creates_all_fail_still_runs_out_of_admission_budget` |
| throttled attempt falls through to the handler | `a_caller_whose_creates_all_fail_still_runs_out_of_admission_budget` |
| bucket never refills | `a_throttled_principal_recovers_at_the_refill_rate` |
| one bucket shared by all principals | `exhausting_one_principal_leaves_every_other_principal_untouched` |
| full principal table fails open | `a_full_principal_table_refuses_rather_than_forgets` |
| `requester_type` folded into the digest | `a_changed_requester_type_keeps_a_principal_in_the_same_quota_namespace` |

The endpoint test is the create that gets refused, not the limiter's arithmetic: the defect it guards against lives in *where* the handler charges, and a unit test of the bucket would pass with the charge in the wrong place. It asserts the API-server request count, because the property is that a throttled attempt reaches the API server not at all.

`cargo fmt --check`, `cargo clippy --workspace --all-targets -- -D warnings`, and `cargo test --workspace` are clean.

### DIFF
--- a/docs/kobe-docs/api/reference.mdx
+++ b/docs/kobe-docs/api/reference.mdx
@@ -278,8 +278,16 @@ When a requested TTL is reduced, `effective_ttl` reports the accepted value.
 | `403` | Missing Sandbox `lease` permission or resource ceiling exceeded |
 | `404` | An allowed SandboxPool does not exist |
 | `409` | Alias is already active |
-| `429` | Sandbox concurrency limit reached |
+| `429` | Sandbox concurrency limit reached, or the per-principal admission rate limit was hit |
 | `503` | Pool configuration or Kubernetes admission cannot be verified |
+
+Admission is rate limited per principal, independently of the concurrency
+limit. The budget is charged per *attempt* — a refused attempt costs the same
+as an accepted one, because refusing one still costs the API server a lease
+create, a contested reservation, and the cleanup that undoes them. A caller
+that has run out of budget gets `429` with a `Retry-After` header, in seconds;
+the concurrency `429` carries no such header. The limit is per API replica, and
+`kobe_sandbox_admission_rate_limited_total` counts every refusal.
 
 ### `GET /v1/sandbox-leases`
 
--- a/src/api/mod.rs
+++ b/src/api/mod.rs
@@ -6,6 +6,7 @@ pub mod sandbox;
 pub mod sandbox_access;
 pub mod sandbox_credentials;
 pub mod sandbox_executions;
+pub mod sandbox_rate_limit;
 pub mod sandbox_streams;
 pub mod sandbox_transport;
 pub mod upgrade;
--- a/src/api/routes.rs
+++ b/src/api/routes.rs
@@ -58,6 +58,11 @@ pub struct AppState<B: ClusterBackend> {
     /// which fans out into dozens of API calls) skip the per-request token
     /// validate + lease GET + kubeconfig Secret read + reqwest client build.
     pub connect_cache: ConnectCache,
+    /// Per-principal admission budget for the Sandbox create path. Shared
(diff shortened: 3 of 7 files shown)


jleni:

PR #141, opened 2026-08-13
chore(sandbox): retire terminal SandboxLease objects

### LINKED ISSUE #123: chore(sandbox): retire terminal SandboxLease objects

emmanuelm41:
Closes #123.

`Released` and `Expired` leases already stop consuming capacity — `consumes_capacity()` is false for both, and #120 releases their quota reservations against verified teardown. This was never a leak. The objects themselves accumulate: one row per Sandbox ever leased, forever. This adds the retention sweep that bounds it.

## Retention default: 7 days (`168h`)

The number is a policy call; the unit is the part that matters. A terminal lease is the audit record of who held what, for how long, and whether cleanup was proven — the thing somebody reads *after* noticing that something went wrong. A window measured in hours would routinely have the evidence gone before the question was asked. One measured in months grows `etcd` without bound holding records nobody will ever read. A week spans a weekend plus the working day it takes to notice, which is the shortest window that still answers the question the record exists to answer.

Configurable via `agentSandbox.leaseRetention` in the chart (`KOBE_SANDBOX_LEASE_RETENTION` for a non-Helm deployment), in the `h`/`m`/`s` grammar `parse_duration` already accepts. There is no `d` unit, which is why the default reads `168h` — teaching `parse_duration` about days would have widened lease TTL parsing at the same time, for no benefit here.

An unusable value (blank, `7d`, `forever`, past the one-year ceiling) falls back to the default rather than to zero or to "sweep nothing": zero deletes records on a typo, and silently disabling the sweep reintroduces the growth this exists to stop. Values below one hour are clamped up, which is not a limit on wanting a short window — it is a guard against `7s` from a dropped character wiping the audit trail with no error anywhere to explain it.

## Quarantined is exempt

**`Quarantined` leases are never retired by this sweep, at any window.** That phase is terminal *and* deliberately still consuming capacity: it is what withholds a slot whose teardown nobody could prove. Sweeping it would hand that slot back on no evidence at all, and delete the record of the unproven teardown in the same call — the exact double-booking the quarantine exists to prevent, now with nothing left to show for it.

Age does not make a quarantine safe to delete. Only a later transition to `Released`/`Expired` does, and the window then runs from *that* transition. The phase test is an exhaustive `match` with no wildcard arm, so a phase added later is a compile error at the one place the decision belongs rather than a silent inclusion.

## How a lease is dated

By its `CleanupVerified=True` condition, written at the terminal transition and nowhere else.

- `creationTimestamp` is the tempting alternative and is wrong: it dates the lease's birth, so a lease that ran for a fortnight would be retired the instant it was released — precisely when its record is most likely to be wanted.
- The `False` form of the same condition dates the moment teardown became *unprovable*, so reading it as an end date would feed the sweep exactly the objects it must never touch.
- A lease with no readable end date is never retired. Keeping it costs one row; deleting it may destroy the record of a lease that ended a minute ago.

Eligibility is re-decided against the **re-read** object, not the listed one. Teardown verification is asynchronous, so a lease can be quarantined in the gap — and the resourceVersion precondition would not catch it, because the controller writes status, so the version being fenced on is already the quarantined one.

## Where it lives

Beside `run_execution_reaper`, not beside the `run_sandbox_admission_reaper` the issue names — no such function exists on this stack. The admission sweep is `reap_abandoned_pending_leases`, which runs at create time only and says so in its own doc comment. The execution reaper is the durable interval in this area, so this follows its shape: runs on every replica, hourly, each delete fenced on UID and resourceVersion so racing replicas produce one delete and a 404/409 for the loser.

Deleting a lease also retires what it owns — its admission reservations, and the internal `ClusterLease` holding a child composition's teardown receipt, both by `ownerReference`. Both are already settled before a clean terminal phase is written, so the sweep never reasons about live capacity; but the whole evidence chain does retire together, which is what the window is buying.

## Verification

`cargo fmt --all`, `cargo clippy --workspace --all-targets -- -D warnings` clean, `cargo test --workspace` green. Chart rendering checked with `helm template`, including the override.

Mutation-checked, five:

| Mutation | Fails |
| --- | --- |
| Sweep includes `Quarantined` | `quarantined_leases_are_never_swept`, `a_lease_quarantined_after_the_list_is_left_alone`, `only_clean_terminal_leases_past_the_window_are_retired` |
| Re-read eligibility fence dropped | `a_lease_quarantined_after_the_list_is_left_alone` |
| `CleanupVerified=False` accepted as an end date | `a_terminal_lease_is_dated_by_its_verified_cleanup` |
| Retention window ignored | `only_clean_terminal_leases_past_the_window_are_retired` |
| Floor on a configured window dropped | `an_unusable_retention_setting_never_shortens_the_window` |

The first mutation found a bad test rather than a bug, and is worth stating plainly: real quarantined leases carry `CleanupVerified=False`, so the original sweep test was spared twice over — by the phase *and* by having no verified end date — and **passed with the phase exemption deleted**. The test now also sweeps a quarantined lease bearing a `True` condition dated a year back, where the phase exemption is the only thing that can save it.

### DIFF
--- a/charts/kobe/templates/deployment.yaml
+++ b/charts/kobe/templates/deployment.yaml
@@ -73,6 +73,12 @@ spec:
             # direct (non-Helm) deployment cannot route around it.
             - name: AGENT_SANDBOX_MODE
               value: {{ .Values.agentSandbox.mode | default "disabled" | quote }}
+            # Retention window for terminal SandboxLease records. Always set,
+            # including in `disabled` mode, so switching the mode on does not
+            # also silently change the retention policy from whatever the
+            # operator wrote here to the built-in default.
+            - name: KOBE_SANDBOX_LEASE_RETENTION
+              value: {{ .Values.agentSandbox.leaseRetention | default "168h" | quote }}
             - name: POOL_SERVICE_ACCOUNT
               value: {{ .Values.poolServiceAccount | default (include "kobe.serviceAccountName" .) | quote }}
             - name: POOL_PUBLISHER_SERVICE_ACCOUNT
--- a/charts/kobe/values.yaml
+++ b/charts/kobe/values.yaml
@@ -98,6 +98,21 @@ containerSecurityContext:
 #             rather than falling back, so it cannot be switched on by accident.
 agentSandbox:
   mode: disabled
+  # How long a SandboxLease that reached a clean terminal phase (`Released` or
+  # `Expired`) is kept before the record is deleted. Those phases stop
+  # consuming capacity when they are written, so this is not about reclaiming
+  # anything — it bounds the one-object-per-Sandbox-ever-leased growth of etcd,
+  # and it costs the audit record of who held what and whether cleanup was
+  # proven. Set it to however long you would want to be able to answer that
+  # question after the fact.
+  #
+  # `Quarantined` leases are NEVER retired by this sweep, at any window. That
+  # phase is what withholds a slot whose teardown nobody could prove; deleting
+  # it would hand the slot back on no evidence and take the record with it.
+  #
+  # Units are h/m/s (there is no `d`). Values under 1h are clamped to 1h, and
+  # anything unparseable falls back to the default rather than to zero.
+  leaseRetention: 168h
 
 clusterPolicies:
   # Restricts the operator's namespace deletion permission (granted
--- a/src/api/sandbox.rs
+++ b/src/api/sandbox.rs
@@ -1838,6 +1838,300 @@ async fn reap_abandoned_pending_leases(
     }
 }
 
+/// How long a lease that reached a clean terminal phase is kept before the
+/// record itself is retired. Seven days, expressed in the units
+/// [`parse_duration`] understands.
+///
+/// The number is a policy call, and the unit is the part that matters. A
+/// terminal lease is the audit record of who held what, for how long, and
+/// whether cleanup was proven — the thing an operator reads *after* somebody
+/// notices something went wrong. A window measured in hours would routinely
+/// have the evidence gone before the question was asked; one measured in months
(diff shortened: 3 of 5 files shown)


emmanuelm41:
Heads-up for review: CI does not run on this PR. `ci.yml` triggers `pull_request` only for `branches: [main, "epic/**"]`, and this targets `feat/sandbox-conformance`, so no checks are reported. Everything in the Verification section above was run locally against this branch. The same gap applies to every PR stacked on a `feat/**` base rather than on `epic/agent-sandbox`; not changing the trigger here, since widening it is a CI-policy call for the whole stack rather than part of this change.

jleni:

PR #143, opened 2026-08-13
test(e2e): expose restart, failure-injection and pty control for Sandbox conformance

### LINKED ISSUE #138: test(e2e): expose restart, failure-injection and pty control for Sandbox conformance

emmanuelm41:
Closes #138.

Three rows of #76's matrix needed the target **disturbed** rather than merely queried, and `hack/e2e.ts` could not do it. Writing those scenarios without the harness would have produced tests that cannot run — worse than not having them, because a skipped scenario reads as a passing one.

The suite still asserts nothing but the public HTTP contract. The disturbance lives on the other side of a process boundary, invoked through `KOBE_SANDBOX_HARNESS`. A suite able to break its own target could also mask a break it did not intend; one reaching into the cluster to check its work could pass while the API was broken.

## The three capabilities

**`restart-operator --wait-for-phase=<stage> --lease=<id>`** waits for one lease to reach a named point, restarts the operator Deployment, and returns only once it is *both serving and leading* again. `rollout status` and `/readyz` only prove the HTTP server is up, and it is up in every replica — the reconcilers run solely in whichever replica holds the `kobe-operator` Lease, acquired afterwards. Acting on the sooner signal drives a cluster whose controllers are not running yet.

Two stages #76 asks for are deliberately **absent**, with the reason in the code: `provisioning`, because nothing in production writes that phase, and `bootstrap`, because the template/warm-pool references on `status.target` are only ever copied from themselves. A stage backed by a marker nobody writes hangs to the timeout and reports a restart that never happened.

**`inject-failure` / `clear-failure`**. `--kind=teardown-unverifiable` revokes `get` on `sandboxclaims` and `clusterleases` — not `delete`. That distinction is the point: a revoked delete stalls in `Releasing` and retries forever, a clean failure; only a read the operator is not permitted to make is the durable uncertainty quarantine exists for. Because RBAC has no deny, a matching rule is **split** so its sibling resources keep every verb they had; a wildcard rule is refused rather than edited, since subtraction cannot narrow `*` and the injection would silently do nothing while reporting success. `--kind=child-api-unreachable` swaps a child API Service's selector for one nothing carries — deleting the StatefulSet would make teardown succeed for the wrong reason.

**`attach-pty`** drives `kobe sandbox attach` through a real terminal. The allocator is `python3 -c 'pty.spawn(...)'` and not the obvious `script(1)`: BSD `script` calls `tcgetattr` on its own stdin, and a harness necessarily feeds keystrokes through a pipe, so it dies with `tcgetattr/ioctl: Operation not supported on socket` before the command starts — observed, not assumed. util-linux tolerates it, so `script` would work in CI and fail on every maintainer's laptop, by producing no output, which is indistinguishable from the keystroke-delivery bug this exists to detect.

Plus **`reap-lease`**: quarantine is terminal, keeps consuming capacity on purpose, and has no exit through the API. Right product behaviour, wrong test behaviour — a scenario proving quarantine happens would starve every scenario after it.

## Six new dual-placement scenarios (twelve runs)

- a restart before readiness resumes the same composition — same claim UID, same Pod UID, same `readyAt`;
- a restart between a retry and its original still runs the command once, asserted through a side effect;
- a restart during teardown still settles the lease once, never quarantining one that had proof;
- an unverifiable teardown withholds capacity instead of releasing it;
- a keystroke typed on a real terminal reaches the workload — typed and expected text differ on purpose, because a pty echoes input;
- Ctrl-C reaches the workload rather than the client.

`both_placements!` now forwards leading attributes, so each carries a doc comment saying what breaks when it does not hold.

## Guard fixes

The suite-shape guard split on `"// Scenarios"`, which is also a substring of any doc comment starting `/// Scenarios`. Writing one silently moved the marker and the guard examined an empty span — it failed safe, but a guard relocatable by prose is not a guard. Now anchored to a line start, and mutation-checked. The scenario floor moved 10 → 17.

## Deliberately not written

No scenario uses `child-api-unreachable`. Severing a child API means nothing under management placement, so the scenario would be single-placement — the exact thing the guard prevents. The capability ships; a scenario for it needs a way to say "this placement only" that does not weaken the matrix.

## Verified / not verified

**Verified locally:** `bun test hack/e2e.test.ts` (17 pass), `--help`, every new subcommand's argument validation, and the pty machinery end to end against a stand-in for `kobe` — keystrokes delivered, Ctrl-C interrupting a running `sleep`, and the assertion failing when the marker never appears. `cargo fmt`, `cargo clippy --workspace --all-targets -- -D warnings`, and `cargo test --workspace` are clean; the suite-shape guards pass.

**Not verified:** every new scenario, and the harness's cluster-facing paths (rollout, leader wait, ClusterRole patch, Service severing, lease reaping). They need a live Kobe endpoint with both a management and a child pool, and none was available. They are `#[ignore]`d and will run for the first time on a real cluster.

## One thing found, and not fixed here

On this branch nothing in production writes `SandboxLeasePhase::Provisioning` or `status.provisioningDeadline` — the only caller of `begin_sandbox_provisioning` is a unit test. `mark_sandbox_ready` requires the deadline, and `transition_sandbox_phase` disallows `Pending → Ready`. A lease created through the HTTP API therefore appears unable to reach `Ready` at all, which would stall *every* scenario in this file at `wait_ready`, old and new. That is a controller change rather than a harness one, and worth its own issue.

### DIFF
--- a/hack/e2e.test.ts
+++ b/hack/e2e.test.ts
@@ -1,6 +1,14 @@
 import { expect, test } from "bun:test";
 
-import { guestImagesForBackend, parseArgs } from "./e2e";
+import {
+  decodeKeystrokes,
+  guestImagesForBackend,
+  LEASE_STAGES,
+  leaseStageReached,
+  parseArgs,
+  ptyCommand,
+  revokeVerb,
+} from "./e2e";
 
 test("k3s only preloads the always-warm k3s guest image", () => {
   const args = parseArgs(["up", "--backend", "k3s"]);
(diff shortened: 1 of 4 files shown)


emmanuelm41:
Heads-up: **this PR gets no CI**, and neither will anything else stacked on `feat/agent-sandbox-integration`.

`ci.yml` triggers on `pull_request: branches: [main, "epic/**"]`, and this targets `feat/agent-sandbox-integration` — so the whole Agent Sandbox stack is accumulating unverified, which is the same failure mode the `epic/**` entry was added to fix.

There is a second, narrower gap: the `paths:` filter lists `hack/**` but not `tests/**` or `justfile`, so a change confined to the conformance suite would skip CI even on a PR into `main`.

Both are one-line fixes in `.github/workflows/ci.yml`; left out of this PR deliberately so a test-harness change does not also carry a CI-trigger change. Everything here was verified locally instead — `cargo fmt`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo test --workspace`, and `bun test hack/e2e.test.ts` are all clean.

PR #183, opened 2026-09-02
fix(api): re-read a Sandbox lease when the release fence loses

jleni:
## What

`DELETE /v1/sandbox-leases/:id` fences its intent patch on the resourceVersion it just read and turned any 409 into `503 Failed to request Sandbox release`, with no retry. The controller writes a fresh lease several times between admission and readiness, so a caller cancelling in that window lost the fence and was told the service was unavailable.

This is what broke the nightly `conformance (k3s)` leg on 2026-08-31 and 2026-09-01: `cancelling_while_provisioning_leaves_nothing_behind` (child placement) panicked with `could not release: HTTP 503`. The PR-path suite does not run that scenario, so the push run on the same commit was green.

## Fix

- On 409, re-read the lease and decide again against the current object, up to five attempts with a short linear backoff. Every early return (already released, releasing, quarantined, intent already recorded, pending-admission cleanup) is made on the fresh read, so a retry never writes intent over newer terminal state.
- A fence that keeps losing is a contended lease, not an outage: answer `409` with reason `conflict_retryable`, the code `extend` already uses.
- Docs: note the fenced retry and the 409 on the release endpoint.

## Tests

- `release_re_reads_when_its_fence_loses_to_a_controller_write`: the mock apiserver bumps the lease on every read; the first patch loses, the second fences on the re-read version and lands. Asserts two reads and the exact fence sequence.
- `release_reports_a_persistently_lost_fence_as_a_retryable_conflict`: the patch always loses; asserts `409 conflict_retryable` after exactly the attempt budget.

The full conformance matrix was dispatched on this branch to exercise the nightly path before merge.

### DIFF
--- a/docs/kobe-docs/api/reference.mdx
+++ b/docs/kobe-docs/api/reference.mdx
@@ -488,6 +488,11 @@ lease returns `204 No Content`. A `Quarantined` lease returns `409 Conflict`:
 cleanup is still uncertain, so its capacity remains unavailable rather than
 being reported as released.
 
+The intent write is fenced on the lease as read, so a controller write that
+lands in between makes the API re-read and try again. If the lease is still
+changing after a few attempts, the API answers `409 Conflict` with reason
+`conflict_retryable`; retry the release.
+
 ### `POST /v1/sandbox-leases/:id/executions`
 
 Reserves the execution before starting its process, so retrying the same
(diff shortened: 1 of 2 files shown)


jleni:
Full nightly-style matrix dispatched on this branch (run 33624401878): `conformance (k3s)` passed the complete sandbox suite, 25/25, including `cancelling_while_provisioning_leaves_nothing_behind`, which failed on main's nightly on 2026-08-31, 09-01 and 09-02. The remaining non-gating legs were cancelled after that verdict to free the single runner for the PR checks. The vcluster leg failed as expected on this branch; that is #184.

PR #209, opened 2026-09-07
chore: bump version to 0.43.1

jleni:
Patch: everything since v0.43.0 is a `fix:`, and only one touches shipped code.

| commit | ships |
|---|---|
| [#207](https://github.com/kunobi-ninja/kobe/pull/207) | `src/controllers/sandbox.rs` — the pre-checkpoint path now reports `child_receipt_does_not_match_precheckpoint` instead of sharing one reason string with the recorded path |
| [#208](https://github.com/kunobi-ninja/kobe/pull/208) | `hack/` only — the admission-policy wait reports its elapsed time |
| [#206](https://github.com/kunobi-ninja/kobe/pull/206) | CI only — release workflow pin |

The quarantine fix ([#205](https://github.com/kunobi-ninja/kobe/pull/205)) is **not** in this release; it merged before the v0.43.0 bump and shipped there, and is already live in int-pro.

So the user-visible change is one diagnostic reason string on one code path. `just bump 0.43.1` moved the workspace, `Chart.yaml` `version` and `appVersion` and the image pins in lockstep; `check-version-consistency.sh` passes.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3365,7 +3365,7 @@ dependencies = [
 
 [[package]]
 name = "kobe-operator"
-version = "0.43.0"
+version = "0.43.1"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -3432,7 +3432,7 @@ dependencies = [
 
 [[package]]
 name = "kobe-runner"
-version = "0.43.0"
+version = "0.43.1"
 dependencies = [
  "base64 0.22.1",
  "clap",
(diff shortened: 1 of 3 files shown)


jleni:
Rebased onto main, which now carries [#210](https://github.com/kunobi-ninja/kobe/pull/210) and [#211](https://github.com/kunobi-ninja/kobe/pull/211). This release is worth more than it was when opened — it now ships two real teardown fixes rather than only the reason-string rename.

The previous `conformance (k3s)` failure was **not** the quarantine bug:

```
[child] exec failed: {"error":"the sandbox runner did not answer","reason":"runner_unreachable"}
    an_execution_returns_the_exact_remote_exit_code ... FAILED
```

`runner_unreachable` is the same infrastructure flake that failed the v0.43.0 release run on a different test. Second occurrence in two days, unrelated to any change here — worth its own investigation rather than a rerun each time.

PR #230, opened 2026-09-11
feat(sandbox): support private image pull secrets

emmanuelm41:
Sandbox pools can now reference administrator-owned, same-namespace image pull Secrets through `template.imagePullSecrets`. Kobe validates the bounded references and projects them into controller-owned Sandbox Pods, enabling private workspace images without exposing service-account or caller-controlled Pod configuration.\n\nValidation: `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`, and `cargo test sandbox --no-fail-fast`.

### DIFF
--- a/charts/kobe/crds/sandboxpools.yaml
+++ b/charts/kobe/crds/sandboxpools.yaml
@@ -350,6 +350,31 @@ spec:
                       type: object
                     maxItems: 64
                     type: array
+                  imagePullSecrets:
+                    description: |-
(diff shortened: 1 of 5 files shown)


emmanuelm41:
Closing: platform RKE2 nodes already route docker.io through the authenticated tenant registry mirror, and Kobe lease profiles route in-lease docker.io pulls through that mirror. Private Docker Hub workspace images need no SandboxPool imagePullSecrets API.

Issue #273, opened 2026-09-16
agent-workspace: no file transfer, and getting files in spends the exec budget


jleni:
Split out of #259, which raised three unrelated problems from the same session.

## What happens

There is no way to put a file into a sandbox or take one out.

Getting in:
- A 2.5 KB script goes in as base64 inside `python3 -c` argv.
- argv above roughly 16 KB is rejected as `command is not acceptable`.
- `--stdin` is refused above its own bound, and deliberately so: it is documented as being for secrets and small inputs, not file transfer.
- A larger overlay took **66 execs in 4 KB chunks**.

Getting out:
- Results come back through `kobe port-forward` to an HTTP server started inside the job, which costs an SSH-agent approval per session.

## Why the chunking is worse than slow

Each lease has a budget of about 200 executions, after which the API returns `HTTP 409 execution_limit_exhausted`. Spending 66 of them on one upload is a third of the budget, and when it runs out **a running job becomes unreachable**: no logs, no cancel. The transfer does not merely cost time, it can strand work that is still running.

## Proposal

```
kobe cp <local> <lease>:<path>
kobe cp <lease>:<path> <local>
```

Not counted against the execution budget, since it is not a command. Failing that, a much higher budget for `agent-workspace` would make the chunking survivable, but it would still be the wrong shape.

Note that `kobe exec --stdin` is not the answer here and should not be widened into one: its bound is intentional, and the refusal message says so.

## Context

Pool `agent-workspace`, CLI/API v0.46.0, observed 2026-09-16. Workload: kache PR mutation testing, about 500 mutants.

jleni:
## iroh is probably the right substrate

Worth recording, because the transport already exists: `src/iroh_transport.rs` (#197) is an optional P2P data plane that today carries the **bytes** of attach and port-forward sessions. The control plane stays REST; iroh only moves payload.

Why it fits this issue specifically:

- **It is not an execution.** The whole injury here is that a transfer spends the ~200-execution budget and can strand a running job. Session bytes do not touch that budget, so riding the session data plane removes the failure rather than raising a limit.
- **The handshake already exists.** `POST /v1/sandbox-leases/{id}/session` performs the same authorization and stream registration as the WebSocket upgrade and returns a one-shot ticket plus the replica's node id. A transfer would reuse it rather than inventing an auth path.
- **QUIC suits bulk data**, which base64 inside argv emphatically does not.
- **End-to-end encrypted**, so a relay observes metadata only.

[...]

PR #283, opened 2026-09-16
feat(workspace): let an agent photograph the desktop

jleni:
## Why

Found while testing the consolidated image on a live lease. The desktop works:

```
Xvfb: 1   x11vnc: 1   websockify: 1
noVNC: HTTP 200
```

And nothing in the image can capture it:

```
$ kobe exec <lease> -- sh -c 'command -v xwd import scrot ffmpeg'
(nothing)
```

So an agent can start a desktop, confirm noVNC is serving, and still have no way to see what is on it short of a human opening a browser through a port-forward. For the workload this image exists for, that is the wrong shape: the agent is the one that needs to look.

## What

`scrot` in the apt layer.

```bash
kobe exec <lease> -- scrot -o /tmp/shot.png
```

Chosen over ImageMagick because it does this one job and pulls in imlib2 rather than a general image toolkit, and over `xwd` because that writes a format needing a second conversion step.

## Not in this PR

Getting the file out. `scp` works today over `ssh kobe-<pool>-<name>` and costs no execution budget, which is the answer for now; kunobi-ninja/kobe#273 covers whether that deserves a first-class command.


jleni:
Superseded by #284. A headless VNC client in the CLI gets the framebuffer from the x11vnc that is already running, so it removes the need for a capture tool in the image instead of adding one. It also sidesteps a blocker this PR does not solve: `kobe exec` cannot carry binary on stdout (2048 bytes in, 3676 out, different hash), so `scrot` to a file still had no clean way back out.

Issue #259, opened 2026-09-16
agent-workspace: a job started from kobe exec dies when the exec returns


jleni:
Narrowed to one problem. This issue originally raised three unrelated things from the same session; the other two now live on their own:

- #272 — the sandbox reports 64 CPUs under an 8-CPU quota
- #273 — no file transfer, and getting files in spends the exec budget

## What happens

```bash
kobe exec <lease> -- bash -lc 'nohup ./job.sh > job.log 2>&1 &'
```

looks fine, but the job is gone once the exec session ends: the script's exit trap fires with status 0 and nothing after the first stage runs.

The workaround is `tmux new-session -d`, and then the job has a controlling terminal, which is its own trap: a test that spawns an interactive shell (`bash -ic 'command -v cc'`) gets stopped by job control and hangs. What actually works is:

```bash
tmux new-session -d "setsid -w bash job.sh </dev/null >log 2>&1"
```

## Resolution

Nothing had to be built. The API has accepted a `detach` flag since executions became durable objects — *"return once reserved rather than waiting for the result"* — and `kobe logs --execution` and `kobe cancel --execution` already address one afterwards. The CLI simply never sent the field, so the entire tmux-plus-setsid dance was a way around a flag that already existed.

#271 exposes it as `kobe exec --detach`.

## Context

Pool `agent-workspace`, CLI/API v0.46.0, observed 2026-09-16. Workload: kache PR mutation testing, `cargo mutants --in-diff … --jobs 2`, about 500 mutants.

PR #288, opened 2026-09-17
feat(workspace): rsync in the image, and say that file transfer works

jleni:
Closes the actionable half of #273.

## What was actually wrong

`scp` and `sftp` have worked over `ssh kobe-<pool>-<name>` all along. Nothing said so, so #259 moved a 2.5 KB script in as base64 inside `python3 -c`, and a larger overlay took **66 executions in 4 KB chunks** — a third of a lease's ~200 budget, where running out leaves a live job unreachable with no logs and no cancel.

I verified the working path on a live lease before writing any of this:

| Direction | Size | Result |
|---|---|---|
| In | 200,000 bytes | `d243fb5649381339` both ends |
| Out | 150,000 bytes | `dc5f54a537c18bc7` both ends |

Byte-exact both ways, four consecutive transfers, **none counted as an execution**: a transfer rides the attach stream, not the execution API. The sandbox is created on first connection, so it works from cold.

## What this adds

**`rsync` in the image.** It is the one transfer shape genuinely missing, because rsync needs the binary at both ends: the caller has it (`/usr/bin/rsync` on macOS) and the sandbox did not. Incremental sync of a source tree is the workload in #259 that started this.

**Documentation**, which is the real defect. A `## Copying files` section on the SSH page with the three commands, the note that a transfer is not an execution, and why this rather than `kobe exec`, whose output is converted to text and corrupts anything binary.

## No CLI wrapper, deliberately

Not `kobe cp`, and not a `kobe rsync`. The line I would draw: **build a client when none exists; do not wrap one that does.**

`rsync` and `scp` are already on the caller's machine and already speak the transport, so a wrapper inherits their flag surface — support all of it and it earns nothing, support a subset and it is a worse `scp`.

That is the opposite of `kobe vnc` in #285, and the difference is measurable rather than a matter of taste:

```
$ which rsync        →  /usr/bin/rsync
$ which vncviewer    →  (nothing)
```

There was no VNC client to wrap, and `kobe exec` could not carry the framebuffer. There is an rsync to use.

## Left open in #273

Whether a pool that declares no SSH capability deserves `kobe cp` at all. That is the only case this does not serve, and it should be settled on its own merits rather than because the issue title asked for a command.


jleni:
Two questions came up on this change: the licence, and whether the client half
of rsync that people actually have on their machine can talk to the copy this
image installs.

**Licence.** rsync is GPL-3.0 (it was GPL-2.0 up to 2.6.x). It arrives here as
a Debian package in a `debian:bookworm-slim` image that already carries bash,
coreutils, git and sed under the same or a similar licence. Installing it
alongside kobe is aggregation, so kobe stays Apache-2.0 and nothing about the
binaries we build changes.

**Interoperability, which was the real risk.** macOS does not ship GNU rsync.
Apple ships `openrsync`, a BSD-licensed reimplementation that reports
`protocol version 29`, while bookworm's rsync 3.2.7 speaks protocol 32. I ran
a bookworm container with an rsync daemon and drove it from a macOS 26 client:

| Case | Result |
| --- | --- |
| Pull, 200 KB binary | sha256 identical |
| Push, 150 KB binary | sha256 identical |
| Delta on a 300 KB file, half unchanged | 149 800 B matched, speedup 1.96 |
| `-az` push | sha256 identical |

The delta line is the one that matters: the rolling-checksum algorithm survives
the version gap, so an incremental source-tree sync does less work on the
second run rather than resending everything. That is the whole reason to prefer
rsync over `scp` here.

One caveat worth knowing rather than documenting: `--info=progress2` does not
exist in openrsync. `--progress`, `--delete`, `--exclude`, `--partial` and
`--checksum` all do, and the commands in the docs use none of the missing ones.

Issue #305, opened 2026-09-17
conformance: child placements take ~98s each, and that is 72% of the job


jleni:
`conformance (k3s)` takes about 28 minutes, and 20 of them are one step. Filed
after #235 turned out to be worth ninety seconds of it.

## Where the time goes

Measured from the step timings of a single `conformance (k3s)` job:

| Step | Duration | Share |
| --- | ---: | ---: |
| Sandbox conformance (management + child) | 20.1 min | 72% |
| Bring up the e2e harness (kind + chart + pools) | 5.7 min | 20% |
| k3s conformance smoke gate | 1.5 min | 5% |
| mise, buildx, checkout, teardown | under 1 min | 3% |

So it is not image builds, not the harness, and not setup. It is the scenarios.

## The part that looks wrong

Within those scenarios, the two placements are not comparable:

```
management sandbox-d25a007e989f: ready at +12.1s
management sandbox-5310b061d7d0: ready at +12.2s

child sandbox-41949b2622d5:      ready at +98.5s
child sandbox-7db1952b1bb1:      ready at +98.5s
child sandbox-9997f5fa52c5:      ready at +98.5s
child sandbox-4ac863c8e3a9:      ready at +98.5s
```

A child placement takes about eight times as long as a management one, and
**four different sandboxes report the same figure to the tenth of a second**.
That is not the spread you get from real provisioning work. The harness polls
`/v1/sandbox-leases/{id}` every 2 seconds (`wait_ready`), so the measurement is
quantised but the underlying event is clearly deterministic — which points at a
fixed interval somewhere on the operator side of the child path rather than at
work that takes 98 seconds to do.

Every scenario that uses child placement pays this, and most scenarios run
against both placements, so it compounds into the 20 minutes above.

## Worth finding out

- What the child readiness path waits on, and whether the ~98s is a reconcile
  requeue, a poll interval, or genuine k3s child boot time.
- Whether that interval can be shortened, or the wait made event-driven, without
  weakening the readiness guarantee it protects.
- Whether scenarios that do not specifically test child placement need to run
  against it at all.

If the ~98s turns out to be real provisioning, this issue is about running fewer
child placements rather than making them faster — but it is worth knowing which
before optimising either.

jleni:
[...]

PR #315, opened 2026-09-18
Revert "perf(sandbox): advance child placements on child-cluster events (#312)"

jleni:
Reverts #312. It makes `conformance (k3s)` fail, and the operator log says why rather than leaving it to inference.

## The evidence

`release_rejects_further_access` fails with the child lease reaching `Quarantined` instead of `Released`, reason `child_receipt_authority_unverified` — *Teardown could not be verified; capacity is withheld*. The controller log for one such lease:

```
07:42:16.505 WARN quarantined  child_receipt_authority_unverified
07:42:16.541 WARN quarantined  (same lease)
07:42:16.587 WARN quarantined
07:42:16.622 WARN quarantined
07:42:16.662 WARN quarantined
07:42:16.757 INFO consuming already-persisted verified teardown receipt
07:42:16.793 INFO child teardown receipt, evidence and footprint absence checkpointed
```

Quarantined five times in 160ms, and 290ms after the first one the evidence it was waiting for lands — for that same lease.

The burst is the signature. Five reconciles in 160ms is the new child-cluster watch firing on SandboxClaim/Sandbox/Pod events; a 5s requeue cannot produce that. Each reconcile runs the receipt check before the authority has persisted its evidence, `authoritative_child_receipt_matches` returns `Ok(None)`, and the lease quarantines.

**The requeue #312 removed was not wasted latency. It was the margin that kept this race unreachable.**

## Why this is not the usual conformance flake

Conformance failed three times on main today from three different infrastructure causes — `runner_unreachable`, `CertificationPending`, and a buildx cache miss. This one is different in kind: a deterministic state-machine outcome, in the exact path the change makes event-driven, with the log showing the ordering. It did not need a second CI sample to establish.

## Why revert rather than fix forward

The fix needs a barrier between the watch-driven reconcile and the evidence write. Where that barrier goes — a guard, a generation check, or making the quarantine retractable — is a design decision for the author of the watch. Landing a guess on a red main while a release waits is the wrong trade.

#311 stays. It passed conformance on its own and also targets #305.

## The process gap

#312 ran with conformance skipped, so the suite that covers child placements never ran against a change whose entire subject is child placements. Re-landing should carry the `ci:conformance` label.

## Also worth a look, separately

The same lease quarantines five times in a row. Whatever the outcome here, that transition looks neither idempotent nor terminal, and it is worth understanding on its own.

## Impact on v0.50.0

[v0.50.0](https://github.com/kunobi-ninja/kobe/releases/tag/v0.50.0) is already published and contains #312. A v0.50.1 should follow this revert.

## Verification

`cargo check --workspace` and `cargo clippy --all-targets` clean.


jleni:
Closing in favour of fixing forward.

The argument in this PR's description — that the fix needs a design decision from the author of the watch — does not survive reading `authoritative_child_receipt_matches`. It returns `Ok(None)` for six different situations, and they are not the same fact:

| | |
| --- | --- |
| `lease.status` absent | not written yet |
| `status.teardown_evidence` absent | **not written yet — this is the race** |
| lease UID empty | cannot tell |
| name/generation disagree | exists and disagrees |
| identity disagrees | exists and disagrees |
| content disagrees | exists and disagrees |

The caller quarantines on all six. Quarantine is the verdict for durable uncertainty, which is right for the last three and wrong for the first three: withholding capacity because something has not been written yet is incorrect whether or not a watch is firing.

So #312 did not introduce this. It removed the 5s requeue that kept the "not yet" case unreachable, and the conflation underneath was always there — it would have surfaced eventually under load.

`status.releasing_at` already stamps when teardown began (added in #282 for the duration metric), so the bound needed to tell "not yet" from "never" exists in the object.

Fixing forward keeps #312's contribution to #305 and removes the latent conflation, instead of reverting one and leaving the other. The real caution here is not who decides — it is that this is the path where the code says *under-counting capacity is recoverable by an operator, silently double-booking a Sandbox host is not*, so the change only relaxes the provably-absent cases and leaves every disagreement quarantining exactly as before.

